PR #213, opened 2024-02-28
[CU] PR template update

narin:
## Description of Change

- Removed ticket number link from top of template since we have ZenHub connections now
- Added checkbox to make sure we check that the Components Integrations workflow has passed before publishing a new version


.github/pull_request_template.md
```diff
@@ -61,4 +61,5 @@ Code reviewer validation:
 If changes warrant a new version [per the versioning guidelines](https://github.com/department-of-veterans-affairs/va-mobile-library#versioning-policy) and the PR is approved and ready to merge:
 - [ ] Merge `main` into branch
 - [ ] Merge branch to `main`
+- [ ] Verify that [Check Component Integrations](https://github.com/department-of-veterans-affairs/va-mobile-library/actions/workflows/check-component-integrations.yml) workflow ran successfully
```
TimRoe, line 64: Did we have this happen more than just this instance of it not liking an absolute vs relative import? If not, can't we can configure linting to check that case so we shouldn't encounter it again? AKA not add another manual step as instead it'd be automated with linting failing to run on the branch preventing merging.

TimRoe (changes requested):
Couple questions about changes

PR #228, opened 2024-03-06
[Feature] Alert - Create Alert Component

TimRoe:
## Description of Change

Created initial pass at Alert component.

- Created Alert component
   - 4 variants with both light/dark mode color schemes
   - Optional header, description, and children (custom react elements) to display as alert content
   - Optional primary/secondary buttons for actions associated with the Alert
   - Split off: Ticket #246 for adding support for expand/collapsible accordion form
- Created a basic Storybook story for testing with Alert
- Button component
   - Updated Button to instead of filling out horizontal width with `alignSelf: 'stretch'` to use `width: '100%'` as it was noted working through Alert that `alignSelf: 'stretch'` will appear bugged (stretch out vertically) if it is nested within a View (or other element) with flexDirection of 'row' and there's space to do so
- Utilities
   - Added new "Base" utility function that handles setting base color scheme (different grays depending on light/dark mode) as a convenience
      - This function is exposed to consuming apps with the `AlertContentColor` function so apps can just call that to correctly set color for any children elements without needing to manually build "base" colors
   - Added new "Spacer" utility component that handles setting spacing between elements in components more simply/explicitly than juggling spacing entirely with margin/padding

No test build, shifted AC for testing in-app to the ticket adding expand/collapse support.

### Screenshots/Video

iOS (similar on Android):

https://github.com/department-of-veterans-affairs/va-mobile-library/assets/8690976/50054f9a-c291-4493-b85c-1affba7f89ce

## Testing

- [x] Tested on iOS 
- [x] Tested on Android 
- [x] Tested on Web

## PR Checklist
Code reviewer validation:
- General
	- [x] Acceptance criteria: 
		- All satisfied _or_
- Code

## Publish

While changes warrant a new version [per the versioning guidelines](https://github.com/department-of-veterans-affairs/va-mobile-library#versioning-policy), given in-app testing was not performed and split to ticket #246 that will be a more appropriate ticket to publish Alert with.

### DIFF
--- a/packages/components/src/components/Alert/Alert.tsx
+++ b/packages/components/src/components/Alert/Alert.tsx
@@ -1,41 +1,48 @@
 import { Colors } from '@department-of-veterans-affairs/mobile-tokens'
 // import { HapticFeedbackTypes } from 'react-native-haptic-feedback'
-import { ScrollView, Text, TextStyle, View, ViewStyle } from 'react-native'
-import React, { FC, RefObject, useEffect, useState } from 'react'
+import { Text, TextStyle, View, ViewStyle } from 'react-native'
+//      ^ ScrollView,
+import React, { FC } from 'react'
+//                ^ , RefObject, useEffect, useState
 
 // import { triggerHaptic } from 'utils/haptics'
 // import { useAutoScrollToElement } from 'utils/hooks'
 
+import { Base, Spacer, useColorScheme } from '../../utils'
 import { Button, ButtonProps, ButtonVariants } from '../Button/Button'
 import { Icon, IconProps } from '../Icon/Icon'
-import { Spacer, useColorScheme } from '../../utils'
+
+/** Convenience function to set children content color correctly with light/dark mode */
+export const AlertContentColor = Base
 
 export type AlertProps = {
   /** Alert variant */
   variant: 'info' | 'success' | 'warning' | 'error'
-  /**  */
+  /** Optional header text */
   header?: string
-  /**  */
+  /** Optional a11y override for header */
   headerA11yLabel?: string
-  /**  */
+  /** Optional description text */
   description?: string
-  /**  */
+  /** Optional a11y override for description */
   descriptionA11yLabel?: string
-  /**  */
+  /** Optional custom content to nest inside Alert
+   * Use AlertContentColor or appropriate component props to set light/dark mode 'base' gray colors */
   children?: React.ReactNode
-  /**  */
+  /** Optional primary action button */
   primaryButton?: ButtonProps
-  /**  */
+  /** Optional secondary action button */
   secondaryButton?: ButtonProps
   /** Optional boolean for determining when to focus on error alert boxes (e.g. onSaveClicked). */
-  focusOnError?: boolean
+  // focusOnError?: boolean
   /** Optional ref for the parent scroll view. Used for scrolling to error alert boxes. */
-  scrollViewRef?: RefObject<ScrollView>
+  // scrollViewRef?: RefObject<ScrollView>
   /** optional testID */
   testId?: string
 }
 
 /**
+ * Work in progress:
  * Displays content in a box styled as an alert
  */
 export const Alert: FC<AlertProps> = ({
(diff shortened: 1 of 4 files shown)


packages/components/src/components/Alert/Alert.tsx
```diff
@@ -0,0 +1,259 @@
+import { Colors } from '@department-of-veterans-affairs/mobile-tokens'
+// import { HapticFeedbackTypes } from 'react-native-haptic-feedback'
+import { Text, TextStyle, View, ViewStyle } from 'react-native'
+//      ^ ScrollView,
+import React, { FC } from 'react'
+//                ^ , RefObject, useEffect, useState
+
+// import { triggerHaptic } from 'utils/haptics'
```
TimRoe, line 8: Added AC to expand/collapse ticket to clean this stuff up since [the Slack thread was a little slow in confirming we don't intend to have a follow-up ticket (pre-MVP) for adding support](https://adhoc.slack.com/archives/C05DQPHUVNZ/p1710874405242539?thread_ts=1709747499.865419&cid=C05DQPHUVNZ).

packages/components/src/utils/index.tsx
```diff
@@ -4,9 +4,18 @@ import {
...
   StyleProp,
+  View,
   ViewStyle,
 } from 'react-native'
-import { useEffect, useState } from 'react'
+import { Colors as TokenColors } from '@department-of-veterans-affairs/mobile-tokens'
+import React, { FC, useEffect, useState } from 'react'
+
+/** Function to prefill base gray colors */
+export function Base() {
```
narin, line 14: This utility function probably won't be needed once we get the semantic tokens/theming in, but for now can we use a more descriptive name for this? Maybe something like `BaseColor`.

packages/components/src/utils/index.tsx
```diff
@@ -66,3 +75,25 @@ export function PressableOpacityStyle(
...
+}) => {
+  return (
+    <View
+      style={{
+        width: horizontal ? size : 'auto',
+        height: !horizontal ? size : 'auto',
+      }}
+    />
+  )
+}
```
narin, line 99: Definitely see how this will be useful moving forward. What do you think about it living in its own file in `components/common` or `components/utils`? Feel like it's a little different than the utils we have so far. 

The organization of this utils folder is probably something we should think about as we add more so `index.tsx`.

narin (commented):
A couple of suggestions but overall looks great.

packages/components/src/utils/index.tsx
TimRoe: Hmmm, I created the function because I wasn't sure we'd be exposing that--I know that would sort things for our components, but would it be implemented in a way that was easy for an app to pull in to apply the same colors to their custom content sent in within the `children`?

If so, maybe should have a TODO to remove this when that comes.

I can rename it to be a bit more descriptive in the meantime.

packages/components/src/utils/index.tsx
TimRoe: I was thinking of it like Icon--its' own component (if very simple one), but not a fully documented one and one I thought made sense to keep hidden away in the utils until the appropriate sizing tokens were established. Although it works just fine without the tokens so it could be made available now.

May be part of the larger question around how we want to expose utils generally--like the useIsScreenReaderEnabled hook I left despite it no longer being used with inline Link going away (for now) because it'd doubtlessly be helpful for apps besides flagship (and technically allow flagship some cleanup too).

Maybe split off a separate ticket to move it either to be the Spacer component or decide if we want a `components/common` potpourri of little things.

TimRoe: Created ticket #257 for this. 

After thinking on it, wrote the ticket to call for a `Spacer/Spacer.tsx` structure as I didn't like the idea of having common/util "components" since Alert is directly using Button and it feels like we could get into a space of splitting hairs around use cases and making judgment calls for a delineation we already have with Icon of "proper" components have doc site documentation while more technical helper/convenience components only appear in Storybook to highlight their existence to technical folks. I think we'll benefit from the the simplicity long term of handling even simple components like any other component (aside from not documenting them fully).

narin (approved):
Approving under the condition that Base util function be renamed to something more descriptive.

Issue #272, opened 2024-04-04
Design Tokens: Create semantic color tokens - loading indicator


jessicawoodin:
**To dos**
- [x] Determine what semantic color tokens are needed for the Loading Indicator component
- [x] Add color tokens to spreadsheet
- [x] Create color tokens in dev environment
- [x] Update fg to foreground
- [x] Create pull request to update css-library version
- [x] Create pull request to update token files
- [x] Add tokens to Figma

Issue #271, opened 2024-04-04
Design Tokens: Create semantic color tokens - alert


jessicawoodin:
**To dos**
- [ ] Determine what semantic color tokens are needed for the Alert component
- [ ] Add color tokens to spreadsheet
- [ ] Create color tokens in dev environment
- [ ] Add tokens to Figma

jessicawoodin:
Ticket no longer needed -- alert tokens will be created with snackbar ticket.

Issue #270, opened 2024-04-04
Design Tokens: Create semantic color tokens - link


jessicawoodin:
**To dos**
- [ ] Determine what semantic color tokens are needed for the Link component
- [ ] Add color tokens to spreadsheet
- [ ] Create color tokens in dev environment
- [ ] Add tokens to Figma

jessicawoodin:
Ticket no longer needed -- all link-related tokens will be created with other components

Issue #179, opened 2024-02-09
Design Tokens: Create semantic color tokens - segmented control


jessicawoodin:
**To dos**
- [x] Determine what semantic color tokens are needed for the Segmented Control component
- [x] Add color tokens to spreadsheet
- [x] Create color tokens in dev environment
- [x] Review with Narin & Tim

PR #276, opened 2024-04-05
[Feature] Update tokens to support css-library 0.5.1

narin:
## Description of Change

The VADS team changed the way they structure their color tokens with css-library 0.5.1 ([PR here](https://github.com/department-of-veterans-affairs/component-library/pull/1074/files#diff-5adf51640c36ace982f86344c0ed73aa56b0e19a321713483ce12933fff6b0dc)), adding an additional level of nesting for their semantic colors in `variables.json`. This PR contains updates to support the new format, clean up figma and JS output, and prepare for semantic token work:
- Updates the transform group to use `cti` in order to support the change in format
- Updates filtering and name clean up for JS output
  - **Note:** Previously we were filtering out the system name (e.g. Vads or Uswds) and removing other words such as "Color" and "System". Undoing this so that we stay consistent with naming with figma and are able to tell the origin of the token. We will need to update the colors in the components accordingly after we've completed semantic tokens
- Fixes figma output with proper typing and aliases (referencing)
- Adds sample for semantic and component tokens for @jessicawoodin 
- Adds new filters for dark and light tokens
- Sorts output by key alphabetically

#### Testing Packages

N/A

### Screenshots/Video

N/A

## Testing

N/A

## PR Checklist
Code reviewer validation:
- General
	- [x] Acceptance criteria: 
		- All satisfied _or_
- Code

## Publish

The tokens will not be published until we've finished the semantic tokens:
- #179 
- #270 
- #271 
- #272 
- #178


.gitignore
```diff
@@ -19,7 +19,6 @@ coverage
 packages/components/.yarn/install-state.gz
 packages/components/.yarn/cache
 packages/tokens/dist/
-packages/tokens/figma/
```
TimRoe, line 22: Was this just removed for testing purposes? The intent for it was so only the [build job](https://github.com/department-of-veterans-affairs/va-mobile-library/actions/workflows/tokens-build.yml) would be updating the figma folder instead of a mix of PRs and automation.

narin: Reverted

packages/tokens/src/tokens/color/component-dark.json
```diff
@@ -0,0 +1,9 @@
+{
+  "button-primary-background-on-dark": {
+    "name": "button-primary-background-on-dark",
+    "value": "{semantic-test-on-dark}",
```
TimRoe, line 4: Is it correct to assume these are just for testing as of this ticket and will no longer have `test` in there with the follow-up tickets?

narin: Yes. Anything in the semantic or component json files will be replaced with non-test tokens in the follow up tickets listed above.

packages/tokens/build.js
```diff
@@ -44,42 +64,80 @@ StyleDictionary.registerFormat({
...
   name: 'rn',
-  transforms: ['name/ti/camel', 'name/color/clean-up', 'color/hex'],
+  transforms: ['name/cti/camel', 'color/hex'],
 })
 
 /** Registering transform group to massage output as desired for figma */
 StyleDictionary.registerTransformGroup({
   name: 'figma',
-  transforms: ['name/ti/camel', 'name/color/clean-up', 'color/hex'],
+  transforms: ['name/cti/kebab', 'color/hex'],
```
TimRoe, line 124: Did we change this to kebab for consistency with VADS/USWDS? If so, perhaps we should also change the tokens themselves to kebab? 

I think we should maintain consistency as otherwise it adds unnecessary cognitive load for designers/engineers to be communicating token names slightly differently.

TimRoe: As a follow-up thought, not sure if that'd be annoying in the code then requiring all token references to wrap kebab-case names in a string or something. To which, if so, I'd push for Figma using camelCase as well and chalking it up as a meaningful web vs mobile differentiation as HTML more naturally using kebab-case while JS/TS doesn't like kebab-case naming.

packages/tokens/build.js
```diff
@@ -44,42 +64,80 @@ StyleDictionary.registerFormat({
...
 })
 
+/**
+ * Utils
+ */
+
+const sortTokensByKey = (obj) => {
```
TimRoe, line 131: Not sure if you encountered issues with this to make it this way (as JS can be finnicky with in-place updating data), but just noting that at face value it seems weird it's instantiating a new object and repopulating it with data instead of the original object after sorting still pointing to the appropriate data for each key.

TimRoe: Nvm, makes sense because JS objects aren't naturally ordered so you can't sort an existing one.

packages/tokens/build.js
narin: I think as long as the names are the same (e.g. `UswdsSystemColorBlueVivid60` in JS and `uswds-system-color-blue-vivid-60` for figma) I'm fine with the difference in casing and I don't think there'll be much cognitive load in communicating. Previously I was filtering out `vads` and `SystemColor` so undoing that keeps the names identical to the original but with different casing. 

@jessicawoodin expressed an interest in keeping the token names the same as VADS for figma because at some point, we're going to have to move over to VA's Figma and we'll be directly referencing their tokens which will be in kebab case, and kebab case just doesn't make sense for JS. VADS will likely want to continue to use kebab case since they're mainly working in CSS.

TimRoe: Not a fan of the inconsistency, but fair play to this part overruling:
> keeping the token names the same as VADS for figma because at some point, we're going to have to move over to VA's Figma

TimRoe (approved):
Looks good.

One side note:
> If any breaking changes, in accordance with the [pre-1.0.0 versioning guidelines](https://github.com/department-of-veterans-affairs/va-mobile-library#versioning-policy): a CU ticket has been created for the VA Mobile App detailing necessary adjustments with the package version that will be published by this ticket

Will be needed at the end of this chain of tickets I think, but not until published--may be helpful to add an AC to the last ticket to be worked as a reminder.

Issue #240, opened 2024-03-13
BUG - Sev-3 - All - Link Icon is not vertically aligned at larger text sizes


narin:
## What happened?

When zoomed in, Icons to the left of Links are not vertically aligned with the first line

Addition: Depending on when this is worked, a similar solution will also be needed on the Alert component which expects similar icon behavior and is initially copying Link's method.
- Moved addition to #246 as these changes were targeted as an escalated bug to be resolved for flagship implementation of the Link component

## Specs:

- Device: Any
- OS: Any

## Steps to Reproduce

1. Increase OS text size to largest size
2. View a Link with an Icon

## Desired behavior

Icon should be vertically centered and aligned with first line of text

## Acceptance Criteria

- [x] Icon is vertically centered and aligned with first line of text
   - [x] Icon remains vertically centered and aligned with first line of text even in the most extreme case of very large text + using preventScaling or maxWidth to keep the icon very small despite the very large text (see image in comments)

## Bug Severity - BE SURE TO ADD THE SEVERITY LABEL
 
See [Bug Tracking](https://department-of-veterans-affairs.github.io/va-mobile-app/docs/QA#issue-severity) for details on severity levels

- Impact: Low 

- 3 - Low

## Screen shot(s) and additional information

<img width="369" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd0RiQXc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--33e6cc63dc905715419d935a8e13736a3d6ccce2/image.png" alt="image.png" />

<details>
  <summary>Full JSON response for services related to issue (expand/collapse)</summary>
  

</details>

## Ticket Checklist
- [ ] Steps to reproduce are defined
- [ ] Desired behavior is added
- [ ] Labels added (front-end, back-end, relevant feature, accessibility, etc)
- [ ] Estimate of 1 added (for front-end tickets)

TimRoe:
Adding more extreme example of this bug (and added an AC for validation case) from flagship's implementation where they prevent scaling on the icon and it looks much worse with large font size:
![image](https://github.com/department-of-veterans-affairs/va-mobile-library/assets/8690976/7bb8a18d-425d-424c-9b6b-f394de7458c0)

PR #288, opened 2024-04-15
[Feature] Alert Collapse/Expand

narin:
## Description of Change

Adds collapse/expand functionality to Alert.
- Added the following props:
  - `exapandable`: determines whether alert is expandable or not
  - `initializeExpanded`: optional boolean that can be passed for when an alert needs to start off as expanded
- Had to restructure the header a bit since we want the title text to be pressable as well in addition to the chevron icon.  This caused some complexity in scenarios where a description is provided but a header is not. Added logic that determines the header text and wraps the header text either in a Pressable or an accessible View based on whether the Alert is collapsible and applies the correct accessibility labels and roles
- Remove unused scrollTo and haptics code
- Tested in mobile app (see video below)
 

#### Testing Packages

- components@0.12.1-beta.2
- components@0.12.1-beta.3
- components@0.12.1-beta.4

### iOS expand/collapse

https://github.com/department-of-veterans-affairs/va-mobile-library/assets/786704/1a909cd6-2fc4-4f59-8c05-3cfaa07b2dd4

### iOS expand/collapse description only

https://github.com/department-of-veterans-affairs/va-mobile-library/assets/786704/11dbb5f2-6a38-4cc7-b9b1-fbe806db50ca

### iOS voiceover

https://github.com/department-of-veterans-affairs/va-mobile-library/assets/786704/cdaac35d-ca32-49a9-aa93-d75e66d08c29

### iOS voiceover – description only

https://github.com/department-of-veterans-affairs/va-mobile-library/assets/786704/122b117a-20e0-407d-9655-7cf339c5af66

### Android

https://github.com/department-of-veterans-affairs/va-mobile-library/assets/786704/05785b71-c982-4d3b-a80e-47d648546830

### Web

https://github.com/department-of-veterans-affairs/va-mobile-library/assets/786704/8bbd415d-2c59-46e4-846c-bda0ba554880

### Web with chevron spacing
image

### VA Mobile App address validation

https://github.com/department-of-veterans-affairs/va-mobile-library/assets/786704/591b3083-a408-487a-afe8-82c95c955275

## Testing

- [x] Tested on iOS 
- [x] Tested on Android 
- [x] Tested on Web
- [x] Tested in VA Mobile App

## PR Checklist
Code reviewer validation:
- General
	- [x] Acceptance criteria: 
		- All satisfied _or_
- Code

## Publish

- [ ] Merge branch to `main`


TimRoe:
> Had to restructure the header a bit since we want the title text to be pressable as well in addition to the chevron icon. This caused some complexity in scenarios where a description is provided but a header is not.

Probably should've been documented in the ticket or Figma, but I'd asked about whether an Alert could be collapsible [in the Q&A thread in Slack](https://adhoc.slack.com/archives/C05DQPHUVNZ/p1709836441322099?thread_ts=1709747499.865419&cid=C05DQPHUVNZ) and Jessica had responded:
> If there is no header, the alert should not be collapsible.

packages/components/src/components/Alert/Alert.tsx
```diff
@@ -29,15 +23,15 @@ export type AlertProps = {
   /** Optional custom content to nest inside Alert
    * Use AlertContentColor or appropriate component props to set light/dark mode 'base' gray colors */
   children?: React.ReactNode
+  /** Optional boolean to determine whether the alert should be expandable */
+  expandable?: boolean
+  /** Optional initial state for expandable alert. Defaults to collapsed */
+  initialExpandedState?: 'collapsed' | 'expanded'
```
TimRoe, line 29: 1. Wonder if this should just be named something like `initializeExpanded` to just be a boolean as well? So then consumers could just drop `initializeExpanded` prop on there more simply than `initialExpandedState='expanded'` to override
2. Don't feel strongly, but perhaps we should use TS to only allow the `initial...` prop if expandable is set. Low stakes since sending it without expandable does nothing, but also slightly better for consumers to not include as a suggestion until after they've already marked it as expandable

packages/components/src/components/Alert/Alert.tsx
```diff
@@ -29,15 +23,15 @@ export type AlertProps = {
   /** Optional custom content to nest inside Alert
    * Use AlertContentColor or appropriate component props to set light/dark mode 'base' gray colors */
   children?: React.ReactNode
+  /** Optional boolean to determine whether the alert should be expandable */
```
TimRoe, line 26: Given it's a boolean, may be clearer doc to just jump straight to something like: 
`True to make the Alert expandable/collapsible`

packages/components/src/components/Alert/Alert.tsx
```diff
@@ -52,41 +46,23 @@ export const Alert: FC<AlertProps> = ({
...
-  // }, [
-  //   variant,
-  //   header,
-  //   description,
-  //   focusOnError,
-  //   scrollRef,
-  //   scrollToAlert,
-  //   scrollViewRef,
-  // ])
+  const [expanded, setExpanded] = useState(
```
TimRoe, line 57: Don't feel strongly since maybe it's clearer this way, but wouldn't this work just as well only using `let expanded = ...` and flipping the variable instead of useState?

packages/components/src/components/Alert/Alert.tsx
```diff
@@ -194,6 +162,52 @@ export const Alert: FC<AlertProps> = ({
     lineHeight: 30,
   }
 
+  const _header = () => {
+    const headerText = header ? (
+      <Text style={headerFont}>{header}</Text>
+    ) : description ? (
```
TimRoe, line 168: As noted in general comment, Q&A with Jessica noted we don't need to support the no-header case for expandable Alerts so this and subsequent logic can be adjusted.

packages/components/src/components/Alert/Alert.tsx
```diff
@@ -29,15 +23,15 @@ export type AlertProps = {
   /** Optional custom content to nest inside Alert
    * Use AlertContentColor or appropriate component props to set light/dark mode 'base' gray colors */
   children?: React.ReactNode
+  /** Optional boolean to determine whether the alert should be expandable */
+  expandable?: boolean
```
TimRoe, line 27: Adding a comment below and thinking more on my general comment of the Jessica Q&A that only header-having Alerts  can be expandable, I wonder if that should be locked down with TS so `expandable` only becomes available as a prop after `header` is set.

packages/components/src/components/Alert/Alert.tsx
```diff
@@ -194,6 +162,52 @@ export const Alert: FC<AlertProps> = ({
...
+        <Pressable
+          onPress={() => setExpanded(!expanded)}
+          accessibilityRole="tab"
+          accessibilityState={{ expanded }}
+          aria-label={a11yLabel}
+          style={{ flexDirection: 'row' }}>
+          <View style={{ flex: 1 }}>{headerText}</View>
+          <View style={iconViewStyle}>
+            <Spacer horizontal />
+            <Icon
```
TimRoe, line 193: Maybe worth checking with Jessica by bumping the [Slack Q&A thread](https://adhoc.slack.com/archives/C05DQPHUVNZ/p1709747499865419) but based on Figma boxes for [expandable Alert examples](https://www.figma.com/file/Zzt8z60hCtdEzXx2GFWghH/branch/DlEO0NFO9TclCmAqQ8giCi/%F0%9F%93%90-Component-Library?type=design&node-id=258%3A2769&mode=design&t=Fxf6YQGYx7hTpdrL-1), it visually appears like the the chevron should also carve out space from the content so that text doesn't appear below the chevron when expanded and would wrap before being out that far.

If that is the case, I think this Icon needs to be moved below to be in the same View with `iconDisplay` below.

TimRoe: [Added a Q&A in the thread](https://adhoc.slack.com/archives/C05DQPHUVNZ/p1713289982748569?thread_ts=1709747499.865419&cid=C05DQPHUVNZ) for this while adding another somewhat related Q&A.

TimRoe (changes requested):
Requesting changes to denote end of first pass with main things being it shouldn't be expandable without a header set and the [Q&A's I added to Slack](https://adhoc.slack.com/archives/C05DQPHUVNZ/p1713289576011039?thread_ts=1709747499.865419&cid=C05DQPHUVNZ) pending input from Jessica.

Also some general suggestions. Did not look beyond the code for the first pass as I expect those things will involve a decent amount of reworking.

Lastly, added a small AC to the ticket from #240 as that is ready to be merged and was done in a targeted way to appease flagship pushing as a necessary bugfix for Link--basically just add [what was done for Link](https://github.com/department-of-veterans-affairs/va-mobile-library/pull/275/files) to the Icon on Alert. Should probably also be applied to the chevron to keep it centered with the text.

packages/components/src/components/Alert/Alert.tsx
narin: 1. Renamed to `initializeExpanded` and changed to boolean
2. Added TS to only allow `initializeExpanded` if `expandable` is true

packages/components/src/components/Alert/Alert.tsx
narin: Added TS logic that requires a header for expandable Alerts

packages/components/src/components/Alert/Alert.tsx
narin: Removed this logic

packages/components/src/components/Alert/Alert.tsx
narin: Added a spacer to the right of the content to achieve this margin when expandable. Considered using flexbox for this, but it would have required me separating the single Pressable into 2 Pressables for the title and chevron. Thought it'd be better for the screenreader to just have the one.

narin:
Made the following suggested changes:
- Rename `initialExpandedState` prop to `initializeExpanded` and changed to boolean
- Add TS union types to require `header` when `expandable` and disallow `initializeExpanded` when `expandable` is falsy
- Remove description header text logic
- Add spacer to right to keep content within expand icon when Alert is expandable
- Add font scaling for icons

Ready for second review.

packages/components/src/components/Alert/Alert.tsx
```diff
@@ -33,13 +32,23 @@ export type AlertProps = {
...
+  | {
+      /** True to make the Alert expandable */
+      expandable: true
+      /** Header text. Required if expandable. Optional if not */
+      header: string
+      /** True if Alert should start expanded. Defaults to false */
+      initializeExpanded?: boolean
+    }
+  | {
+      expandable?: false
```
TimRoe, line 47: Didn't check, but may also need doc on expandable/header here to correctly show helper text when leveraging the component since by default expandable would be false.

narin: Good call. Added.

packages/components/src/components/Alert/Alert.tsx
TimRoe: I think we are going to want to unify this in some capacity with the logic below. Can check with Jessica, but playing with the component currently the touch target for expand/collapse is currently only the header text (and trailing space) + the chevron. It doesn't include the alert icon (info/check/etc.) or space that's visually part of the collapsed Alert but not close enough to the text/chevron, leaving a fairly large area to the left end and the border on top/right/bottom sides that doesn't toggle the expand/collapse behavior. 

Intuitively, I think tapping anywhere on the collapsed Alert should expand it--and similarly tapping anywhere along the top portion should collapse when expanded. Perhaps even further: when expanded tapping anywhere on the Alert that isn't an internal pressable (Buttons, children components with onPress) should also collapse it.

Basically: wondering if the entire Alert should be wrapped in the Pressable and we just have no onPress for non-expandable versions (and correspondingly hide the expand/collapse a11y stuff behind expandable).

packages/components/src/components/Alert/Alert.tsx
```diff
@@ -194,6 +180,43 @@ export const Alert: FC<AlertProps> = ({
...
+
+    /**
+     * Wrap header text and expand icon in Pressable if the Alert is expandable
+     * Otherwise wrap in View with accessibility props
+     */
+    if (expandable) {
+      return (
+        <Pressable
+          onPress={() => setExpanded(!expanded)}
+          accessibilityRole="tab"
```
TimRoe, line 195: 1. Swap to `role` for the new/more-web-like accessibility props
2. Is this the right role? I don't see anything better looking at the list, but doesn't really seem like a tab per se... maybe should confirm with Jessica/Brea?

packages/components/src/components/Alert/Alert.tsx
```diff
@@ -194,6 +180,43 @@ export const Alert: FC<AlertProps> = ({
...
+          accessibilityState={{ expanded }}
+          aria-label={a11yLabel}
+          style={{ flexDirection: 'row' }}>
+          <View style={{ flex: 1 }}>{headerText}</View>
+          <View style={iconViewStyle}>
+            <Spacer horizontal />
+            <Icon
+              fill={contentColor}
+              width={Sizing._16}
```
TimRoe, line 204: Per the other item I added to [the Q&A thread](https://adhoc.slack.com/archives/C05DQPHUVNZ/p1709747499865419), chevron should have a maxWidth of 24.

packages/components/src/components/Alert/Alert.tsx
```diff
@@ -229,31 +252,32 @@ export const Alert: FC<AlertProps> = ({
       <View style={{ flexDirection: 'row' }}>
         {iconDisplay}
```
TimRoe, line 253: Per the item I added to [the Q&A thread](https://adhoc.slack.com/archives/C05DQPHUVNZ/p1709747499865419), iconDisplay via either the icon itself or iconProps should have preventScaling true.

packages/components/src/components/Alert/Alert.tsx
narin: This was done based on CollapsibleAlert. Looks like Brea had already checked off on `tab` being the correct role. Although I'm now seeing that it's just the header that should be a `tab` and the entire View should be a `tablist`. Corresponding ticket here: https://github.com/department-of-veterans-affairs/va-mobile-app/issues/4365#issuecomment-1363362436

packages/components/src/components/Alert/Alert.tsx
```diff
@@ -229,31 +252,32 @@ export const Alert: FC<AlertProps> = ({
...
+                  <View
+                    accessible={true}
+                    aria-label={descriptionA11yLabel || description}>
+                    <Text style={descriptionFont}>{description}</Text>
+                  </View>
+                ) : null}
+                {description && children ? <Spacer /> : null}
+                {children}
+              </View>
+              {expandable && <Spacer horizontal size={36} />}
```
TimRoe, line 270: If retaining doing it this way, shouldn't the Spacer be 26? 

10 Spacer + 16 wide chevron and then needing to scale 26->34 with font scaling.

packages/components/src/components/Alert/Alert.tsx
narin: Confirmed with Jessica that the entire Alert should be pressable when it's collapsed. Will have to rework the structure so that header is separate from content.

packages/components/src/components/Alert/Alert.tsx
narin: Updated

packages/components/src/components/Alert/Alert.tsx
narin: Added preventScaling to iconDisplay above.

packages/components/src/components/Alert/Alert.tsx
narin: Updated to 26. Are you saying I need to add logic to set a max width for scaling? I suppose another approach could be to add another invisible Icon with the same props as the chevron, but seems a little hacky.

packages/components/src/components/Alert/Alert.tsx
TimRoe: Alternatively, perhaps we [use `hitSlop`](https://reactnative.dev/docs/touchablewithoutfeedback#hitslop) with a Rect object to fill out the padding to be pressable? Doing it that way, we could also pull the Chevron out (to use flexbox like the alert icon) and ensure the entire "header" is pressable when open or closed. Not sure if that might get weird around font scaling with the hitSlop though.

packages/components/src/components/Alert/Alert.tsx
TimRoe: > Are you saying I need to add logic to set a max width for scaling?

If we want to mimic what the chevron is doing when font scaling comes into play, we'd need to match it growing but only a certain amount otherwise the description/children wrapping will get thrown off.

I think we'd be best suited to avoid using a Spacer here to account for the icon. From that other comment chain, I think either:
- The entire Alert should be Pressable (we can move the chevron icon to be in the View row flex structure down here) or
- We should use hitSlop to pad out the pressable area regardless of icon presents (again we can move the chevron icon to be in the View row flex structure down here)

narin: Added `hitSlop` to expand coverage of the touchable area, but ran into a weird issue when moving the expand icon out of the Pressable. The icon itself wasn't responding to touches, but everywhere else within the hitSlop area was. Couldn't find too much about it on Google except that sometimes hitSlop doesn't work well with SVGs/icons, but strangely it works fine on the left Info icon. See video:


https://github.com/department-of-veterans-affairs/va-mobile-library/assets/786704/2226da49-be34-4360-955b-314af1a34a0f



Ended up just adding an invisible icon I've called `spacerIcon` which is guaranteed match the width of the expand icon at any scale. I tried using a Spacer instead and calculating the proper width with fontScale but it doesn't quite work since the icon and spacer scale at different rates due to the maxWidth.

packages/components/src/components/Alert/Alert.tsx
```diff
@@ -162,6 +167,34 @@ export const Alert: FC<AlertProps> = ({
...
+  /**
+   * When an alert is expandable, the content should have additional padding on
+   * the right to appear within the expandable icon. Since the expandable icon
+   * has a maxWidth, this hidden icon matches the spacing of the icon insteading
+   * instead of adding a <Spacer /> with a calculated value.
+   */
+  const spacerIcon = (
+    <View style={iconViewStyle} aria-hidden>
+      <Spacer horizontal />
+      <Icon {...expandIconProps} fill={backgroundColor} />
```
TimRoe, line 194: Not sure if it matters performance-wise, but wonder if the fill should be `'transparent'`? Not sure if under the hood that would make it not actually render at all (but hopefully respect size) vs. rendering the same color as below.

packages/components/src/components/Alert/Alert.tsx
```diff
@@ -194,6 +211,44 @@ export const Alert: FC<AlertProps> = ({
     lineHeight: 30,
   }
 
+  const _header = () => {
+    const headerText = header ? <Text style={headerFont}>{header}</Text> : null
```
TimRoe, line 215: Seems like it behaves fine (in iOS at least, didn't check Android) with the screen reader, but don't like that when header is undefined (e.g. valid non-expandable version) we're returning a null wrapped in `<View accessible={true} aria-label={a11yLabel} role="heading">`.

I think these header checks are probably a byproduct of the prior description-in-header functionality.

We should probably just open this function with:
`if (!header) return null`
and clean up these conditionals a bit to be sure no awkwardness could happen around returning a valid view with no content in it when we don't have a header.

packages/components/src/components/Alert/Alert.tsx
```diff
@@ -170,13 +167,33 @@ export const Alert: FC<AlertProps> = ({
     </View>
```
TimRoe, line 167: ^^ Looks like we lost `preventScaling` on the alert Icon so it's growing with font changes.

TimRoe: Was confused seeing the touchable area not reaching the left edge and the hitSlop math adding up, only when realizing I had my font scaled up that it was due to the icon growing. Should remedy that problem once preventScaling is back.

packages/components/src/components/Alert/Alert.tsx
TimRoe: Weird--but also something I think I may have run into once upon a time way back in my prior job of touchable weirdness with SVGs.

Unfortunate, but I think the solution makes sense given the odd circumstances.

TimRoe (changes requested):
Looking better. A few minor things:
- preventScaling on alert icon
- Cleanup of the `_header` function a bit
- Suggestion that maybe 'transparent' or 'none' fill on the spacerIcon might be more performant as it may potentially not actual bother fully rendering where I think giving it a color (even the same as below) might

packages/components/src/components/Alert/Alert.tsx
narin: Returning `null` if no header and cleaned up `headerText` and `a11yLabel` logic accordingly.

packages/components/src/components/Alert/Alert.tsx
narin: Re-added `preventScaling`. Good catch.

packages/components/src/components/Alert/Alert.tsx
narin: Updated fill color to `'none'`.

packages/components/src/components/Alert/Alert.tsx
```diff
@@ -194,6 +211,46 @@ export const Alert: FC<AlertProps> = ({
...
+
+    /**
+     * Wrap header text and expand icon in Pressable if the Alert is expandable
+     * Otherwise wrap in View with accessibility props
+     */
+    if (expandable) {
+      return (
+        <Pressable
+          onPress={() => setExpanded(!expanded)}
+          accessibilityRole="tab"
```
TimRoe, line 236:
```suggestion
          role="tab"
```

packages/components/src/components/Alert/Alert.tsx
```diff
@@ -194,6 +211,46 @@ export const Alert: FC<AlertProps> = ({
...
+          accessibilityState={{ expanded }}
```
TimRoe, line 237:
```suggestion
          aria-expanded={{ expanded }}
```
(not sure if same formatting for expanded side)

packages/components/src/components/Alert/Alert.tsx
```diff
@@ -225,35 +282,39 @@ export const Alert: FC<AlertProps> = ({
   }
 
   return (
-    <View style={contentBox} testID={testId}>
+    <View
+      style={contentBox}
+      testID={testId}
+      accessibilityRole={expandable ? 'tablist' : 'none'}>
```
TimRoe, line 288:
```suggestion
      role={expandable ? 'tablist' : 'none'}>
```

TimRoe (approved):
Approved.

Added 3 last minor suggestions to convert to the more web-like RN a11y prop names (accessibilityRole->role, accessibilityState->aria-expanded) since I assume (didn't check) Storybook web is complaining about the old versions being deprecated.

Issue #279, opened 2024-04-09
Jest Matchers - Enhance Accessibility Unit Tests


TimRoe:
## Description 
Split from #134.

It was discovered during #134 that a more proper way of testing the a11yValue override was available using [Jest Matchers](https://callstack.github.io/react-native-testing-library/docs/jest-matchers), but it appeared to require additional Jest configuration that was not already setup and our configuration didn't readily match up to the Jest [guidance for adding Jest Matchers](https://callstack.github.io/react-native-testing-library/docs/getting-started#jest-matchers). Since it was discovered mostly after getting the accessibility tests to work somewhat correctly via brute force, this ticket was split off to handle the necessary Jest configuration changes to enable this more proper way of validating the accessibility (and other helpful additions to the standard data matching Jest offers).

## Acceptance Criteria
  

-  [ ] Update Jest configuration to leverage the enhanced Jest Matchers
-  [ ] Fix any linting errors associated with new eslint extension
-  [ ] Update the `a11y tests` set of Link and Alert tests with the improved Jest Matcher for a11yValue
-  [ ] If possible, also improve the a11y-Label/Hint to actually read values instead of just checking prop values

narin:
Tried to add Jest matchers for Alert accessibility tests and found that the new eslint plugin would require changes to all of our unit tests. Adjusted this AC to account for that, renamed ticket, and bumped estimate to 8 points.

CC: @kellylein

PR #301, opened 2024-04-29
[Feature] Alert unit tests

narin:
## Description of Change

Adds the following unit tests for Alert:
- Check for rendering of header, description, children, buttons
- Verify colors for each variant in light and dark mode
- Expandable variants
- Analytics passthroughs

This PR also renames the `testId` prop in Alert to `testID` to be consistent the other components and fixes one of the Button tests to use queryByText instead of the async findByText.

Note: Moved accessibility tests to #279 since we'll want to configure jest matchers

#### Testing Packages

N/A

### Screenshots/Video

image

## Testing

N/A, they're unit tests

## PR Checklist
Code reviewer validation:
- General
	- [x] Acceptance criteria: 
		- All satisfied _or_
- Code

## Publish

- [ ] Merge branch to `main`

### DIFF
--- a/packages/components/src/components/Alert/Alert.test.tsx
+++ b/packages/components/src/components/Alert/Alert.test.tsx
@@ -258,13 +258,13 @@ describe('Button', () => {
           </Alert>,
         )
 
-        // Only header should be visible before press
         expect(component.queryByText('Header text')).toBeTruthy()
         expect(component.queryByText('Description text')).toBeTruthy()
         expect(component.queryByText('Sample children content')).toBeTruthy()
         expect(component.queryByText('Primary test button')).toBeTruthy()
         expect(component.queryByText('Secondary test button')).toBeTruthy()
 
+        // Only header should be visible after press
         fireEvent.press(component.getByRole('tab'))
         expect(component.queryByText('Header text')).toBeTruthy()
         expect(component.queryByText('Description text')).toBeNull()
--- a/packages/components/src/components/Alert/Alert.test.tsx
+++ b/packages/components/src/components/Alert/Alert.test.tsx
@@ -133,8 +133,8 @@ describe('Alert', () => {
         const { backgroundColor, borderLeftColor } = getContentStyle(component)
         expect(backgroundColor).toBe('#ecf3ec')
         expect(borderLeftColor).toBe('#008817')
-        expect(getIconName(component)).toBe('Check')
         expect(getTextColor(component)).toBe('#3d4551')
+        expect(getIconName(component)).toBe('Check')
       })
     })
 
--- a/packages/components/src/components/Alert/Alert.test.tsx
+++ b/packages/components/src/components/Alert/Alert.test.tsx
@@ -18,34 +18,41 @@ jest.mock('react-native/Libraries/Utilities/useColorScheme', () => {
   }
 })
 
+// Test props
+const headerText = 'Header text'
+const descriptionText = 'Description text'
+const testID = 'testAlert'
+
+const children = (
+  <View>
+    <Text>Sample children content</Text>
+  </View>
+)
+
+const primaryButtonProps = {
+  label: 'Primary test button',
+  onPress: () => console.log('primary press'),
+  testID: 'primaryButton',
+}
+
+const secondaryButtonProps = {
+  label: 'Secondary test button',
+  onPress: () => console.log('secondary press'),
+  testID: 'secondaryButton',
+}
+
+// Helper utils
+const getContentStyle = (element: RenderAPI) =>
+  element.getByTestId(testID).props.style
+
+const getTextColor = (element: RenderAPI) =>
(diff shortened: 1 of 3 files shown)


packages/components/src/components/Alert/Alert.test.tsx
```diff
@@ -0,0 +1,305 @@
...
+const getContentStyle = (element: RenderAPI) =>
+  element.getByTestId(testID).props.style
+
+const getTextColor = (element: RenderAPI) =>
+  element.getByText(headerText).props.style.color
+
+const getIconName = (element: RenderAPI) =>
+  element.root.findByType(Icon).props.name
+
+describe('Button', () => {
```
TimRoe, line 54:
```suggestion
describe('Alert', () => {
```

packages/components/src/components/Button/Button.test.tsx
```diff
@@ -45,8 +45,8 @@ describe('Button', () => {
       expect(onPressSpy).toBeCalled()
     })
 
-    it('should render label', async () => {
```
TimRoe, line 48: Is there any harm in having things async even when they don't need to be?

With Link opted to have most async because I was getting confusing errors without it for some tests.

packages/components/src/components/Alert/Alert.test.tsx
```diff
@@ -0,0 +1,305 @@
...
+        mockedColorScheme.mockImplementation(() => 'light')
+
+        component = render(
+          <Alert variant="error" header={headerText} testID={testID} />,
+        )
+
+        const { backgroundColor, borderLeftColor } = getContentStyle(component)
+        expect(backgroundColor).toBe('#f8dfe2')
+        expect(borderLeftColor).toBe('#b50909')
+        expect(getTextColor(component)).toBe('#3d4551')
```
TimRoe, line 200: Very nitpicky, but it's a little weird that for all variants the light/dark mode `expect`s swap the order of TextColor/IconName.

packages/components/src/components/Alert/Alert.test.tsx
```diff
@@ -0,0 +1,305 @@
...
+            description={descriptionText}
+            expandable
+            primaryButton={primaryButtonProps}
+            secondaryButton={secondaryButtonProps}>
+            {children}
+          </Alert>,
+        )
+      })
+
+      it('should initially render header only', async () => {
```
TimRoe, line 236:
```suggestion
      it('should initially render header only then render everything after press', async () => {
```
or something to that effect, doesn't adequately describe this test kind of doing double duty.

packages/components/src/components/Alert/Alert.test.tsx
```diff
@@ -0,0 +1,305 @@
...
+        expect(component.queryByText('Header text')).toBeTruthy()
+        expect(component.queryByText('Description text')).toBeTruthy()
+        expect(component.queryByText('Sample children content')).toBeTruthy()
+        expect(component.queryByText('Primary test button')).toBeTruthy()
+        expect(component.queryByText('Secondary test button')).toBeTruthy()
+      })
+    })
+
+    describe('Initialize expanded', () => {
+      it('should render expanded', () => {
```
TimRoe, line 254: Similar comment to above.

TimRoe (changes requested):
Dropping request changes to denote end of first pass--mostly just a few minor things for direct comments.

Non-direct comment: doesn't test accessibility. Seems like at a minimum should check header/description-a11yLabel overrides. Maybe should also check the icon scaling for the main icon and the chevron with its' unique behavior if we can do that around prior knowledge it renders everything at like 200% or whatever.

packages/components/src/components/Button/Button.test.tsx
narin: I don't think so but we should avoid adding it if there's no async functions within. I think the errors might stem from some of the queries are async functions (e.g. findBy*) and some are not (queryBy*). When we add the jest matchers eslint plugin I think it provides some guidance on which ones should be used.

packages/components/src/components/Alert/Alert.test.tsx
narin: Fixed

narin:
Moved accessibility tests to be a part of #279. Ready for another round of review.

packages/components/src/components/Alert/Alert.test.tsx
```diff
@@ -0,0 +1,310 @@
...
+            expandable
+            primaryButton={primaryButtonProps}
+            secondaryButton={secondaryButtonProps}>
+            {children}
+          </Alert>,
+        )
+      })
+
+      it('should initially render header only', async () => {
+        // Only header should be visible before press
```
TimRoe, line 237:
```suggestion

```

TimRoe (approved):

PR #307, opened 2024-05-02
[Tests] Update to use React Native Testing Library Jest Matchers and add accessibility tests

narin:
## Description of Change

Adds React Native Testing Library's Jest matchers as per [setup instructions](https://callstack.github.io/react-native-testing-library/docs/jest-matchers#setup). These matchers provide us with helpful matchers, make tests more readable, and also introduces RNTL specific linting to help us adhere to best practices. This required some updates to our jest and TypeScript setup, as well as changes to all of our tests. Most notable changes:

1. Usage of `screen` for queries instead of assigning the result of the `render` to a variable like we've been doing with `component` (enforced by [testing-library/prefer-screen-queries](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/prefer-screen-queries.md))
2. `render` is not allowed in `beforeEach()` (enforced by [testing-library/no-render-in-lifecycle](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/no-render-in-lifecycle.md)). Added `render()`'s to each test.
3. Replace `toBeTruthy` and `toBeNull` tests with more readable `toBeOnTheScreen` and `not.toBeOnTheScreen`
4. Replace existing accessibility tests in Link with accessibility queries such as `getByLabel` and `getByHintText` or use `hasAccessibilityValue` matcher
5. Add accessibility tests to SegmentedControl, Button, and Alert
6. Replace existing style tests with `toHaveStyle` matcher
7. Replace any color tokens with hex values until color tokens are finalized
8. Wrap Link component in ComponentWrapper. Translated `aria-valuetext` wasn't coming through for usage with `hasAccessibilityValue`

### Screenshots/Video

Screenshot 2024-05-02 at 12 21 22 PM

## Testing

N/A – Updates to tests only

## PR Checklist
Code reviewer validation:
- General
	- [x] Acceptance criteria: 
		- All satisfied _or_
- Code

## Publish

- [ ] Merge branch to `main`


packages/components/tsconfig.json
```diff
@@ -1,5 +1,7 @@
 {
   "compilerOptions": {
+    "module": "NodeNext",
+    "moduleResolution": "NodeNext",
```
TimRoe, line 4: These are already present (commented out) down in the "Modules" section if we were continuing to roll with the structured form of key/values.

packages/components/src/components/SegmentedControl/SegmentedControl.test.tsx
```diff
@@ -27,26 +20,31 @@ describe('SegmentedControl', () => {
   })
 
   const labels = ['segment 0', 'segment 1']
+  const a11yLabels = ['segment 0 a11y label', 'segment 1 a11y label']
```
TimRoe, line 23: Wonder if this should only override one or the other to validate the fallback case of using `labels` when there's a falsy value in the array.

packages/components/src/components/SegmentedControl/SegmentedControl.test.tsx
```diff
@@ -56,65 +54,94 @@ describe('SegmentedControl', () => {
     }
   }
 
-  beforeEach(() => {
```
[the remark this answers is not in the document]
narin: `beforeEach` was calling `initializeTestInstance` which calls `render` inside of it which goes against the [testing-library/no-render-in-lifecycle](https://github.com/testing-library/eslint-plugin-testing-library/blob/main/docs/rules/no-render-in-lifecycle.md) linter rule

packages/components/src/components/Alert/Alert.test.tsx
```diff
@@ -18,292 +18,326 @@ jest.mock('react-native/Libraries/Utilities/useColorScheme', () => {
...
-      })
+    it('should render only the header after being collapsed', () => {
+      render(
+        <Alert
+          variant="error"
+          expandable
+          initializeExpanded={true}
+          {...commonProps}
+        />,
+      )
```
TimRoe, line 310: Did this test need to render everything again or would it still be on screen in collapsed form?  Curious since the render for the prior test is outside the test and doesn't change anything. Maybe this is also bumping up against the linter though.

packages/components/src/components/Link/Link.test.tsx
```diff
@@ -43,57 +42,53 @@ describe('Link', () => {
   const defaultProps: LinkProps = {
```
TimRoe, line 42: Maybe worth renaming `commonProps` for consistency with other tests?

packages/components/src/components/Link/Link.test.tsx
```diff
@@ -424,28 +411,30 @@ describe('Link', () => {
...
+      expect(screen.getByLabelText('a11yLabel override')).toBeOnTheScreen()
     })
 
     it('includes a11yHint', async () => {
-      expect(component.root.props.accessibilityHint).toBe('a11yHint override')
+      render(<Link {...defaultProps} />)
+      expect(screen.getByHintText('a11yHint override')).toBeOnTheScreen()
     })
 
     it('includes a11yValue', async () => {
```
TimRoe, line 432: Mentioned in our 1:1 but probably not very clearly, I think we want a second a11yValue test for poking at sending a custom one. 
> /** Optional a11yValue for "[position #] of [list total #]" or a custom value descriptive string */
  a11yValue?: a11yValue | string

So basically a test rendering with defaultProps + a11yValue sent as a string (AKA they have a custom formatted string description and didn't want to leverage our translation based on indexes). It was done this way as [accessibilityValue is rather flexible in use conditions](https://reactnative.dev/docs/accessibility#accessibilityvalue).

TimRoe (changes requested):
Looks mostly good. Left a couple question/suggestion comments, but main primarily requesting changes for:
1. Adding additional Link a11y test to validate sending a custom string for `a11yValue`
2. Adjusting the Segmented Control a11yLabels const to `[undefined, 'segment 1 a11y label']` and validating that the a11yLabel for the first segment falls back to `segment 0` to validate the fallback behavior of using the label itself without override present for that segment's index

packages/components/src/components/Alert/Alert.test.tsx
narin: Yes, the test would fail with `'render' method has not been called` if there hasn't been a "fresh" render called prior to the test. I suspect the previous test didn't fail since render is called right before, even if it wasn't within the it(). Moved the render to be within.

packages/components/src/components/Link/Link.test.tsx
narin: Renamed

narin:
- Added additional Link a11yValue test for custom string
- Updated SegmentedControl a11yLabel test to account for fallback value
- Renamed `defaultProps` to `commonProps` in Link

TimRoe (approved):

PR #309, opened 2024-05-07
[Tests] Add unit tests for OSFunctions

TimRoe:
## Description of Change

Added unit test suite for OSFunctions.tsx file fully validating both utility functions in the file. 

Additionally made a slight tweak to OSFunctions.tsx itself to convert our `isIOS` convenience constant into a function so it queries the Platform.OS value in real time--which'll never change for actual devices so there is no functional change, but allows taking updates for unit tests to validate platform specific functionality.

As a miscellaneous note: the tests are complicated somewhat by the fact that `Alert.alert` calls OS functionality and thus does not actually appear in Jest [per this StackOverflow answer](https://stackoverflow.com/a/61448169) so it's not possible to perform UI-based inspection/interaction on them.

## Testing

Validated unit tests pass as committed and do not pass flipping values that would expect to cause failures.

Checked in iOS/Android that finding directions still works as intended using Apple/Google Maps to confirm `isIOS` change resulted in no functional change. Omitting Changelog label from this change as it is a trivial tweak and the main thrust of this ticket entails no publishable changes.

- [x] Tested on iOS 
- [x] Tested on Android 
- ~~Tested on Web~~ N/A

## PR Checklist
Code reviewer validation:
- General
	- [x] Acceptance criteria: 
		- All satisfied _or_
- Code

## Publish

Changes no not warrant a new version [per the versioning guidelines](https://github.com/department-of-veterans-affairs/va-mobile-library#versioning-policy).


packages/components/src/utils/OSfunctions.test.tsx
```diff
@@ -0,0 +1,294 @@
...
+        cancel: 'Cancel override',
+        confirm: 'Confirm override',
+      },
+    }
+
+    afterEach(() => {
+      jest.resetAllMocks()
+    })
+
+    it('should call RN Linking with no RN Alert prompt on non-URL variety', () => {
```
narin, line 185: Should we rename this "tel variety" or "non-HTTP variety"? Test description is a little confusing to see "non-URL" then see the url prop used.

narin (approved):
Just one nit regarding a test name but looks great otherwise.

packages/components/src/utils/OSfunctions.test.tsx
TimRoe: Updated to `non-HTTP`.

PR #328, opened 2024-05-21
[Bug] Storybook - Fix Code Snippets and Custom Docs Template

TimRoe:
## Description of Change

Fixed issue with "Show code" dropdown not appearing in the Docs page. It's not entirely clear what fixed it--could've been the version updates I did, but it may have been fixed with Storybook 7 to work with custom doc templates (while simultaneously breaking automatic doc, reversing the issue from the prior version).

After fixing that, Button's doc was broken and wouldn't show all the possible props--this was fixed by getting rid of custom TypeScript generation.

Also after moving to the custom doc template, the controllability of props in the table broke. It appears that was due to using ArgsTable instead of Controls.

Lastly, removed the workaround links to the doc pages in the description block now that the intended method with the custom doc template is working again.

## Testing

Verified everything working as expected in Web Storybook. Spot checked iOS/Android work as before. 

Did notice one very minor bug in Web Storybook where the "Show code" dropdown for the Alert component changes to "No code available" toggling to whichever light/dark mode you didn't initially load into. Toggling it back or refreshing the page fixes it. Removing the Buttons from the Storybook component appeared to fix this so I think it is related to the fact that the Button props go from Primary to Base which changes the properties being sent and changes the code preview which confuses Storybook so I'm not sure it can be fixed as clearly the component's actual behavior swapping those is more important.

- [x] Tested on iOS 
- [x] Tested on Android 
- [x] Tested on Web

## PR Checklist
Code reviewer validation:
- General
	- [x] Acceptance criteria: 
		- All satisfied _or_
- Code

## Publish

Changes do not warrant a new version [per the versioning guidelines](https://github.com/department-of-veterans-affairs/va-mobile-library#versioning-policy) as they only involve Storybook, devDependencies, and removing comments in actual components which is non-functional.


packages/components/.storybook/web/main.ts
```diff
@@ -21,9 +20,6 @@ module.exports = {
   },
   framework: '@storybook/react-webpack5',
   options: { builder: { useSWC: true } }, // improves build performance
-  typescript: {
-    reactDocgen: 'react-docgen',
```
TimRoe, line 25: With Storybook 7, TypeScript is natively supported so this shouldn't be needed. This was screwing up Button's Doc page Controls panel (possibly due to the way we're currently using an enum for setting buttonType values) and I didn't notice any issues removing it.

packages/components/.storybook/native/storybook.requires.ts
```diff
@@ -1,6 +1,10 @@
 /* do not change this file, it is auto generated by storybook. */
```
TimRoe, line 1: Highlighting that all changes in this file happened automatically either from the storybook update process or building after update.

narin (approved):
Looks good! Glad we stumbled upon some solutions here.

Issue #255, opened 2024-03-20
DS - Add missing icons from updated Icon Library


narin:
## Description 
We need to add some icons to the icon library because they are mobile-specific so won't come from VADS.

Dependent on [approval from Design System Council](https://design.va.gov/about/contributing-to-the-design-system/suggest-an-addition-or-update) for adding icons. ([Draft](https://docs.google.com/document/d/1O3R77LCv4Xk-uQqmdc-ZIbCQN0LQSxkLtQC4ZLzyp7s/edit#heading=h.6npd3hx2etin))

AC:   
- [ ] Work with VADS to add 2 Material icons (`inbox` and `refresh`) to their sprite within VADS repo
     - Submitted [request](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2865) to VADS on 5/23/24
     - Ticket was updated from 7 icons to 2 icons because:
          - VADS recently added `tty` and `phone_iphone`
          - VADS already had `local_shipping` in progress
          - [VADS declined to add](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/issues/2865#issuecomment-2135894634) `circle` and `crop_square`. We will need to add [these two icons](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd2dlQlE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b7dce188ec2668fc39dee1b12163e3c2a8f27900/Add%20to%20VADS.zip) in the mobile app repo instead.

- [x] Add [7 checkbox/radio icons](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBMW9RQkE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--7be722d35bd852e901f84cdf581e9c4cd0303fba/Form%20fields.zip) to the mobile library
     - `check_box_error` was removed after discussion between Jessica and Narin -- `check_box` will be used instead
[...]

PR #329, opened 2024-05-21
[Feature] Add script to extract SVGs from VADS sprite sheet

narin:
## Description of Change

Adds a script that takes VADS sprite sheet and extracts individual icons from it. Having this script allows us to keep the latest version of `component-library` as a dev dependency, but extracts only the icons instead of including the entire 3mb package. We will have to run this any time VADS's sprite.svg changes.

Also adds the latest set of extracted icons to the repo.

Tested on all platforms by bringing in the accessibility_new icon that was extract from the sprite sheet.

#### Testing Packages

0.5.1-alpha.3

#### Android

Screenshot 2024-05-21 at 12 59 55 PM

## Testing

- [x] Tested on iOS 
- [x] Tested on Android 
- [x] Tested on Web

## PR Checklist
Code reviewer validation:
- General
	- [x] Acceptance criteria: 
		- All satisfied _or_
- Code

## Publish

- [ ] Merge branch to `main`


packages/assets/extract-svgs.js
```diff
@@ -0,0 +1,28 @@
...
+ * Extracts icons from VADS icon sprite as single svgs
+ * Adapted from https://gist.github.com/mhmdnsr/18343cceb12858c8b148ef31d1e5a32b
+ */
+
+import * as fs from 'fs'
+import { parse } from 'svg-parser'
+import { toHtml } from 'hast-util-to-html'
+
+fs.readFile(
+  '../../node_modules/@department-of-veterans-affairs/component-library/dist/img/sprite.svg',
```
TimRoe, line 11: Is your local client out of sync or something here? It doesn't appear that the `@department-of-veterans-affairs/component-library` is a dependency of this or any of our packages so it seems like this file shouldn't exist.

TimRoe:
Not certain it's a problem and even if it is not sure we can do anything about it from our side, but the way GitHub is displaying some of the new icons has me concerned about potential invisible whitespace between the SVG viewBox and the size of the actual vector drawn image* which would cause unexpected downsizing when trying to use them in the Icon component. Just wanted to document the concern.

*Based on inconsistent gaps between visible size and the box around it--looking back at [an old PR I know the icons are good for](https://github.com/department-of-veterans-affairs/va-mobile-library/pull/181/files) the padding appears consistent regardless of icon dimensions while icons like `accessibility_new.svg` and `accessible_forward.svg` do not relative to others in the PR.

packages/assets/extract-svgs.js
```diff
@@ -0,0 +1,28 @@
...
+  'utf8',
+  function (err, contents) {
+    const parsed = parse(contents)
+    const symbols = parsed.children[0].children
+    let length = 0
+    symbols.forEach((symbol) => {
+      const name = symbol.properties.id
+      symbol.tagName = 'svg'
+      let newIcon = toHtml(symbol)
+      fs.writeFile(`icons/vads/${name}.svg`, newIcon, () => {
```
TimRoe, line 21: Maybe a question we need to loop Jessica in on, but should we try to de-duplicate here with the existing icons we have? Removing name conflicts from either ours or VADS seems maybe easier with the script than on the component side where we either drop icons on the floor or need separate names for them.

TimRoe: Looking through the icons, this would be somewhat problematic with `home` since they only have one and that would make the nav bar icon change in unintended ways (not just go filled to outline, but a totally different house).

TimRoe (changes requested):
- Branch missing the `@department-of-veterans-affairs/component-library` devDependency
- Question on duplication between our icons and VADS

packages/assets/extract-svgs.js
TimRoe: Per stand-up, sounds like we intend to mostly clean up our icons and will have an updated `home` icon that is an outline form of VADS/Material icons. So we do not need to worry about automatic de-duplicating as we'll handle it soonTM on follow-up tickets.

narin:
Added `component-library` to devDependencies. De-duplication and removal of icons was discussed in stand-up as mentioned and will be handled in #255.

TimRoe (approved):

PR #399, opened 2024-07-16
[Spike] Snackbar proof of concept

TimRoe:
## Description of Change

Created a proof of concept for the Snackbar component that works in both Storybook and flagship. This ticket mostly lays some groundwork, significant refinement will be made on implementation ticket for Snackbar including creating the custom display and getting into specifics where this PR leaves fairly general properties copied from flagship.

#### Testing Packages

- [0.20.4-alpha.0](https://www.npmjs.com/package/@department-of-veterans-affairs/mobile-component-library/v/0.20.4-alpha.0)

### Screenshots/Video

Storybook (with 3 second duration for testing):

https://github.com/user-attachments/assets/55fb1cca-795f-4e3d-a72b-25fd6fadc6a9

Flagship (with 3 second duration for testing):

https://github.com/user-attachments/assets/8f7a30ea-7214-42bc-920b-2e62c8daefcb

## Testing

Validated it worked as expected in iOS in Storybook. Validated it worked as expected in flagship (with minor fixes in node_modules from alpha build that were retroactively made on the branch but no new alpha was published).

Current implementation did not work in Web Storybook. That will be deferred to later tickets. In theory, it should be possible as the react-native-toast-notifications package documentation includes a web demonstration of it working. It may require remedial work specifically for Web Storybook though. Worst case scenario, we can just drop the custom Snackbar UI in stories and not have it "pop up" since Storybook is for documentation purposes and that is likely sufficient.

- [x] Tested on iOS 
- [x] Tested on Android 
- [x] Tested on Web
   - Checking off, but it did not work in Web

## PR Checklist
Code reviewer validation:
- General
	- [x] PR is linked to ticket(s)
	- [x] PR has `changelog` label applied if it's to be included in the changelog
	- [x] Acceptance criteria: 
		- All satisfied _or_
		- Documented reason for not being performed _or_
		- Split to separate ticket and ticket is linked by relevant AC(s)
	- [x] Above PR sections adequately filled out
	- [x] If any breaking changes, in accordance with the [pre-1.0.0 versioning guidelines](https://github.com/department-of-veterans-affairs/va-mobile-library#versioning-policy): a CU ticket has been created for the VA Mobile App detailing necessary adjustments with the package version that will be published by this ticket
- Code
	- [x] Tests are included if appropriate (or split to separate ticket)
	- [x] New functions have proper TSDoc annotations

## Publish

[...]

PR #414, opened 2024-07-25
[Feature] Create Snackbar component

TimRoe:
## Description of Change

Updated the Snackbar component proof of concept into a functional custom snackbar meeting design specs from Figma/doc site as much as possible.

Three tickets split from this effort due to proving more complicated than initially expected:
1. A ticket to handle setting the Snackbar offset from the bottom
   - `offset` can only be set at the SnackbarProvider level which is just initialized at the app start--we need a mechanism to be able to update the `offset` property on SnackbarProvider from within the Snackbar component that gets called for display logic
   - #415 
2. A ticket to handle the Snackbar being announced to the screen reader, but not taking focus
    - Ostensibly [announceForAccessibility](https://reactnative.dev/docs/accessibilityinfo#announceforaccessibility) should be able to do this, but putting in a useEffect in the component was not working as intended--it'd only announce for accessibility if multiple Snackbars were opened and would occur when dismissing one of the later ones instead of on open. This potentially entails a similar handling as above of having the logic for announcing the Snackbar at the SnackbarProvider level (always "rendered") while only triggering it when the Snackbar display appears
    - #416
3. A ticket to look into [a flagship-level bug with Snackbars](https://github.com/department-of-veterans-affairs/va-mobile-app/issues/5055)
   - [Draft testing noted](https://adhoc.slack.com/archives/C02N65QRRB7/p1722279055856109) this was a problem in Secure Messaging and Chris' notes suggest the fix likely lies within the Snackbar so we should make an attempt to resolve it as part of the DS Snackbar component to remedy the bug in flagship and prevent similar problems for future component consumers
   - #417

#### Testing Packages

- [0.21.1-alpha.0](https://www.npmjs.com/package/@department-of-veterans-affairs/mobile-component-library/v/0.21.1-alpha.0)

### Screenshots/Video

Storybook (iOS):

https://github.com/user-attachments/assets/569f9429-bb0b-409f-91f3-5e80758a05fc

App (Android):

https://github.com/user-attachments/assets/8a9f8f66-4ea8-4dc1-b1c1-a754378ffd62

## Testing

Sample testing code added within `customOnPress` in Link component:
```
    const snackbarText = 'Message moved to Custom Folder'
    snackbar.hideAll()
    snackbar.show(snackbarText, {
      data: {
        isError: true,
        messageA11y: 'a11y override',
        onActionPressed: () => {
          null
        },
      },
    })
```

Attached videos didn't include, but also tested with varying font sizes/message lengths for wrapping.

- [x] Tested on iOS 
- [x] Tested on Android 
- [x] Tested on Web
   - Did not test in Web, presumably still not working like the proof of concept; see PR #399 for additional details
   - Bumped Snackbar Storybook ticket 3->5 to account for that and this ticket not implementing any Storybook for testing because of how Snackbar functions (just added to the custom button press for Link)

## PR Checklist
Code reviewer validation:
- General
	- [x] Acceptance criteria: 
		- All satisfied _or_
- Code

## Publish

While changes do warrant a new version [per the versioning guidelines](https://github.com/department-of-veterans-affairs/va-mobile-library#versioning-policy), the Snackbar has follow-up work delaying a new version to a later ticket.


packages/components/src/components/Snackbar/Snackbar.tsx
```diff
@@ -42,20 +66,231 @@ type SnackbarProviderProps = {
...
+export const CloseSnackbar = () => {
+  snackbar.hideAll() // Remove any existing snackbars
+}
+
+/**
+ * Convenience handling function to close snackbar(s) on navigation
+ */
+export const CloseSnackbarOnNavigation = () => {
+  CloseSnackbar()
+}
```
TimRoe, line 296: Not sure about these, especially the latter two. During tech discovery I thought they'd add something, but I'm not sure they do.

At some point I thought they'd be useful for analytics, but I don't think the Snackbar actually needs analytics. If we add an analytics passthrough to these, an app may as well just make their own versions and do the trivial `hideAll()` and `show()` themselves. The action button takes custom onPress logic that they can bake analytics into. Analytics on the "Dismiss" button doesn't really make sense--presumably every Snackbar that appears is dismissed, either by the user or switching screens. I also thought maybe there'd be some nice abstraction to mitigate breaking changes using these vs direct calls, but as things evolved I don't think there's anything that typing on `snackbar` can't handle.

Maybe we keep them (or just ShowSnackbar) on principle solely for slight convenience to a future app that doesn't care to have analytics on Snackbars to have a function call vs littering individual `snackbar` calls all over?

packages/components/src/components/Snackbar/Snackbar.tsx
```diff
@@ -1,30 +1,54 @@
...
   }
 }
 
+/**
+ * Structured to allow modification of `react-native-toast-notifications` ToastOptions
+ * type, but Snackbar component locks down all non-data options presently
+ *
+ * In the future, this may change (e.g. allowing non-indefinite `duration`)
+ */
+type modifyToastOptions = snackbarData
```
narin, line 42: Will this be modified at all in the 3 follow-up of tickets? If not, I feel like it  should be removed until this future change is needed, but maybe I'm missing something.

packages/components/src/components/Snackbar/Snackbar.tsx
```diff
@@ -48,17 +72,222 @@ type SnackbarProviderProps = {
...
+
+const SnackbarButton: FC<SnackbarButtonProps> = ({ text, onPress }) => {
+  const theme = useTheme()
+
+  const helperTextBold: TextStyle = {
+    fontFamily: 'SourceSansPro-Bold',
+    fontSize: 16,
+    lineHeight: 22,
+  }
+  const typography = { helperTextBold }
```
narin, line 118: Is this const needed? Seems like it's only used once and could just do `...helperTextBold` below.

packages/components/src/components/Snackbar/Snackbar.tsx
```diff
@@ -48,17 +72,222 @@ type SnackbarProviderProps = {
...
+  //   AccessibilityInfo.announceForAccessibility('Test announce two')
+
+  // useEffect(() => a11yAnnounce, [a11yAnnounce])
+
+  const helperText: TextStyle = {
+    fontFamily: 'SourceSansPro-Regular',
+    fontSize: 16,
+    lineHeight: 22,
+  }
+  const typography = { helperText }
```
narin, line 154: Same comment as above

packages/components/src/components/Snackbar/Snackbar.tsx
narin: Agreed that the third seems needed as it's the same as `CloseSnackbar()`.

I could see both Show and Close being useful as users might be expecting `CloseSnackbar` with the existence of `ShowSnackbar`.

As an aside, is exposing `snackbar` globally required for this to function? If not, maybe it would make more sense for users to only expose the `ShowSnackbar` and `CloseSnackbar` functions which they import.

narin (approved):
A couple nits and clarifying questions. Overall looks good.

packages/components/src/components/Snackbar/Snackbar.tsx
TimRoe: No, just making it structured the same way it'll be when we have the typography tokens--but guess maybe that's taking an unnecessary step since helperTextBold probably won't exist as a typography token anyway.

I can remove it.

packages/components/src/components/Snackbar/Snackbar.tsx
TimRoe: No, I don't expect this would change as part of MVP or currently planned post-MVP changes. 

I retained it because it was picking apart the library typing in a way that may not necessarily be intuitive when revisiting later. Maybe that's fine since it was only mentioned in passing the possibility of allowing controlling the duration or other aspects so we may never end up doing it, but I liked retaining the breadcrumb that the `data` prop is for our custom data while there are other Toast-library level values that could potentially be updated here outside `data` if we allow them in the future. It lingering also leaves a record of why it's called with the nested `data` object since it could be simplified, but then would result in breaking changes if we ever did shift to allowing other props here in the future.

packages/components/src/components/Snackbar/Snackbar.tsx
TimRoe: > Agreed that the third seems unneeded as it's the same as `CloseSnackbar()`.

Yeah, I mainly separated it as it matches flagship's having this function and calling attention to close on navigation as we can't inherently force apps to close Snackbars on navigation.

> As an aside, is exposing `snackbar` globally required for this to function?

Yes. Without the global var `snackbar` set, the `show` and `hideAll` functions aren't set to anything to control behavior.

We could conceivably package the component up differently--I think flagship [implemented it globally](https://www.npmjs.com/package/react-native-toast-notifications?activeTab=readme#--how-to-call-toast-outside-react-components) so that it could be tied into redux to update the bottom offset although I'm not that's even actually needed. The [standard implementation](https://www.npmjs.com/package/react-native-toast-notifications?activeTab=readme#usage) of the library uses a global wrapping component + hook in every place a Toast is called so it can only be called via react-native components. Seemed prudent to minimize the invasiveness for flagship and generally have a more powerful/robust implementation for future apps to maintain the global availability.

The global var `snackbar` does the heavy lifting, these functions are purely for convenience. `ShowSnackbar` actually does something in it enforcing closing any existing Snackbars first, but the `Close-` versions are just wrappers for `snackbar.hideAll()` that a properly configured app should be able to call anywhere.

narin: I do like the standard implementation of having the `useToast` hook. If we can confirm that the app wouldn't need to use the global for redux, I feel like just having either a `useSnackbar` hook OR importing `ShowSnackbar` and `CloseSnackbar` is cleaner than giving them the option to use the global `snackbar` too.

TimRoe: @narin and I discussed in Slack and we have added a note to ticket #415 to look into shifting to use the [standard implementation](https://www.npmjs.com/package/react-native-toast-notifications?activeTab=readme#usage) of the toast library which may simplify some things for other follow-up tickets. The main con being it complicating the implementation for flagship, but this would be a one-time downside for otherwise more standard react(-native) functional implementation of the Snackbar component.

If everything pans out, global implementation can be explored post-MVP if/when the need arises.

PR #421, opened 2024-08-02
[Feature] Add Snackbar to Storybook

TimRoe:
## Description of Change

Added storybook story for the Snackbar component. There is a single story demonstrating a fairly standard case of a successful scenario that includes action button to "Undo" which can easily be modified (in web) to have: 
- `isError: true` to change the icon and action button to "Try again"
- Remove `onActionPressed` to remove the action button

In web, the story displays the Snackbar visual logic flat. 

In iOS/Android, the same story instead has a custom render logic which displays a Button and pressing the button brings up the Snackbar using actual handling logic.

### Screenshots/Video

Web:

https://github.com/user-attachments/assets/96f1ec20-6383-4d85-baef-1e4b39e4caee

iOS (Android behaves similar):

https://github.com/user-attachments/assets/c91cb88d-14b6-4731-be1c-d3f077bc71c6

## Testing

Validated iOS/Android and web behaved as expected.

One minor bug lingers in web--the message text does not wrap and instead expands off the visible display into the void. It is unclear why this is the case as it works fine in iOS/Android and ostensibly should in web too since the message text is styled to flexShrink so the container should grow until it hits the bounds confining the Snackbar, then the message text should stop growing horizontally and expand vertically to wrap... but it doesn't. This seems like a bug with `react-native-web`, storybook, or a behavior difference between `react-native` and CSS flexbox. I have adjusted the default from "Message moved to Custom Folder" to "Message moved to Test Folder" to mitigate this issue appearing with default values (AKA Storybook "Small mobile" preset) so it will only appear modifying the message text in storybook to be longer. Since web is just for demonstration, it didn't seem worth making additional adjustments within the Snackbar component itself attempting to resolve when it works for supported platforms.

- [x] Tested on iOS 
- [x] Tested on Android 
- [x] Tested on Web

## PR Checklist
Code reviewer validation:
- General
	- [x] Acceptance criteria: 
		- All satisfied _or_
- Code

## Publish

Changes do not warrant a new version [per the versioning guidelines](https://github.com/department-of-veterans-affairs/va-mobile-library#versioning-policy).

### DIFF
--- a/packages/components/src/components/Snackbar/Snackbar.tsx
+++ b/packages/components/src/components/Snackbar/Snackbar.tsx
@@ -49,29 +49,6 @@ export type SnackbarType = Omit<ToastType, 'show' | 'update'> & {
   ) => string
 }
 
-//
-// TODO: May need to move the following doc to the custom render to show in Storybook
-//
-
-/**
- * To use SnackbarProvider, your app must have a global.d.ts which you can copy the following into:
- * ```
- * type SnackbarType = import('@department-of-veterans-affairs/mobile-component-library').SnackbarType
- * // eslint-disable-next-line no-var
- * declare var snackbar: SnackbarType
- * ```
- * then add SnackbarProvider in App.tsx (or similar foundational file) level with your app rendering:
- * ```
- * <>
- *   <App />
- *   <SnackbarProvider />
- * </>
- * ```
- *
- * This config will allow it to be called anywhere including outside React components.
- *
- * The Snackbar remains open indefinitely. App configuration should ensure it is dismissed on navigation.
- */
 export const SnackbarProvider: React.FC = () => {
   // const forceUpdate = useReducer(x => x + 1, 0)[1]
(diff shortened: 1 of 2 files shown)


packages/components/src/components/Snackbar/Snackbar.tsx
```diff
@@ -221,7 +217,7 @@ export const Snackbar: FC<SnackbarProps> = (toast) => {
 
   const buttonContainer: ViewProps = {
     style: {
-      flex: 0,
+      flexGrow: 0,
```
TimRoe, line 220: This was changed because it seems otherwise react-native-web (or web flexbox generally) does not respect the flex 0 and `iconAndMessageContainer` just ate all the available space, even when it didn't need it at all. This appeared to result in no functional change in iOS/Android.

narin (approved):
Looks good. Nice job finding the workarounds for web.

PR #423, opened 2024-08-07
[Feature] Snackbar – Add mechanism for adjusting bottom offset

narin:
Refactor Snackbar to use hook/provider pattern instead of global and support offset.
- Added `offset` to `Snackbar` data props which can now be passed along with `.show()`
- Removed `snackbar` from `global.d.ts` and any references to it
- Separated `SnackbarProvider` into a separate file and modified to allow `children` prop as per [documentation](https://github.com/arnnis/react-native-toast-notifications/#usage) 
  - `SnackbarProvider` is a wrapper of `ToastProvider` and has its own context to set and maintain `offset` state
- Created `useSnackbar` hook which is a wrapper of `useToast`
  -  `show()` method adjusts `offset` in `SnackbarProvider` context before calling `toast.show()`
  - If `offset` is provided, sets to custom value before showing, if not, will reset to default
- Removed show/close convenience functions since they were only relevant for the global approach 
- Added Storybook story, separating Default and Custom Offset stories

**Notes:**
- You'll notice I separated `useSnackbar` into its own hook file in a new `hooks/` folder. I think our `utils/index.tsx` is getting large enough to point now where we might want to break them up by category, e.g. `hooks/colors.tsx`, `hooks/accessibility.tsx` and have hooks that may be specific to a component be its own file. If we're all onboard with this pattern, I can create a follow-up ticket to split up the rest of our hooks.

#### Testing Packages

- 0.21.1-alpha.0
- 0.21.1-alpha.1
- 0.21.1-alpha.2
- 0.21.1-alpha.3
- 0.21.1-alpha.5
- 0.21.1-alpha.6
- 0.21.1-alpha.7
- 0.21.1-alpha.8

#### Storybook (iOS)

https://github.com/user-attachments/assets/0da3ee1e-0fd5-424c-9bca-0dc6f4820ad8

#### Storybook (Android)

https://github.com/user-attachments/assets/8ea3b053-8a23-4d75-ba5e-133c82430302

#### App (iOS)

https://github.com/user-attachments/assets/98a6d795-02a0-41ee-ab10-2c703676da58

#### App (Android)

https://github.com/user-attachments/assets/26f06ae1-89e8-4827-a33c-051ad83b1e78

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 
- [ ] Tested on Web

## PR Checklist
Code reviewer validation:
- General
	- [x] Acceptance criteria: 
		- All satisfied _or_
- Code

## Publish

- [ ] Merge branch to `main`


packages/components/src/components/Snackbar/Snackbar.stories.tsx
```diff
@@ -34,27 +35,48 @@ export default meta
...
+      isError: false,
+      messageA11y: 'Message moved to Custom Folder with a11y',
+      onActionPressed: () => console.log('Action pressed'),
+    },
+  },
 }
 
-export const _Snackbar: Story = {
-  render: Platform.OS !== 'web' ? mobileComponentRenderer : undefined, // Render Snackbar flat in web
+export const __CustomOffset: Story = {
```
TimRoe, line 71: Not sure about this being a separate story. Looking in Web, it's identical which is confusing as a demonstration tool.

Not sure if we can restrict this story to only appearing in mobile, otherwise we may be fine just having the Default story default a custom offset and updating as needed when testing on mobile devices.

packages/components/src/components/Snackbar/Snackbar.tsx
```diff
@@ -30,6 +30,8 @@ type snackbarData = {
     messageA11y?: string
     /** action button onPress logic for "Try again" (isError=true) or "Undo" button */
     onActionPressed?: () => void
+    /** offset from bottom of screen. defaults to 50 in SnackbarProvider */
+    offset?: number
```
TimRoe, line 34: Nitpicky, but suggest moving up to be alphabetical.

packages/components/src/components/Snackbar/Snackbar.tsx
```diff
@@ -11,16 +11,16 @@ import {
...
-import Toast, { ToastType } from 'react-native-toast-notifications'
-import ToastContainer from 'react-native-toast-notifications'
 
 import { Icon, IconProps } from '../Icon/Icon'
 import { Spacer } from '../Spacer/Spacer'
 import { useTheme } from '../../utils'
 
-// let snackbarOffset: number = 10
+// TODO: Replace with global setting
+export const DEFAULT_OFFSET: number = 50
```
TimRoe, line 23: Potentially should add `SNACKBAR_` or maybe even `COMPONENTS_SNACKBAR_` or something since if we're defining it flat I think it could also conflict any global variables defined by an app AKA would be safest to be named obnoxiously specifically to prevent any conflict.

packages/components/src/components/Snackbar/Snackbar.tsx
```diff
@@ -155,23 +127,16 @@ export const Snackbar: FC<SnackbarProps> = (toast) => {
     _20: 20,
     _24: 24,
   }
```
TimRoe, line 129: Can't comment on the appropriate lines since they weren't edited, but lines 89-105 should be updated to document the updated explanation for how to implement the Snackbar within an app.

packages/components/src/index.tsx
```diff
@@ -13,13 +13,9 @@ export { Icon } from './components/Icon/Icon'
...
-  ShowSnackbar,
-  SnackbarProvider,
-  type SnackbarType,
-} from './components/Snackbar/Snackbar'
+export { SnackbarProvider } from './components/Snackbar/SnackbarProvider'
 export { Spacer } from './components/Spacer/Spacer'
 
 // Export consumer available utilities here so they are exported through npm
 export { useTheme } from './utils'
+export { useSnackbar } from './hooks/useSnackbar'
```
TimRoe, line 21: Nitpicky: alphabetical order 😃

packages/components/src/components/Snackbar/Snackbar.tsx
```diff
@@ -39,7 +41,7 @@ type snackbarData = {
  *
  * In the future, this may change (e.g. allowing non-indefinite `duration`)
  */
-type modifyToastOptions = snackbarData
+export type modifyToastOptions = snackbarData
```
TimRoe, line 44: Instead of exporting, I think this can just be deleted. And `SnackbarType` below deleted as it's no longer used without the global.

packages/components/src/components/Snackbar/Snackbar.tsx
```diff
@@ -39,7 +41,7 @@ type snackbarData = {
...
 
 export type SnackbarType = Omit<ToastType, 'show' | 'update'> & {
```
TimRoe, line 46: Think this can be deleted as it was used just for defining the global `snackbar` type.

packages/components/src/components/Snackbar/Snackbar.tsx
```diff
@@ -11,16 +11,16 @@ import {
   useWindowDimensions,
```
TimRoe, line 11: Can't comment on the appropriate line, but should be able to remove the `ts-nocheck` from the top of the file now, correct?

packages/components/src/components/Snackbar/Snackbar.tsx
TimRoe: Also, leaving the offset as 50 was just random(ly approximately above the bottom bar in Storybook) since I knew another ticket would follow. I think the default offset should probably be 0 or something more thoughtful. Maybe slightly more than 0 because I think with the keyboard open it then is right against the top of it? Perhaps a Jessica call there, but maybe ensuring a minimum 4 or 8 pixels off the keyboard would be good.

Not sure if we should inherently factor in OS. The offsets in flagship look to be: 
- iOS
   - 25
   - 94 w/ Nav
- Android
   - 0
   - 66 w/ Nav

Not sure if we should have any iOS/Android differences though and leave it to the consumer to manipulate as appropriate from default.

Potentially relevant note: in flagship they are using SafeAreaView which is an iOS thing, but I did not include it because it seemed like the library was handling well enough to keep it on screen and it seemed superfluous. That behavior may have altered what flagship is doing with the numbers above.

packages/components/src/components/Snackbar/Snackbar.stories.tsx
narin: Hid this on web by adding a filter that will hide any stories on web that have a `'mobile-only'` tag

packages/components/src/components/Snackbar/Snackbar.tsx
narin: Removed

packages/components/src/components/Snackbar/Snackbar.tsx
narin: Moved

packages/components/src/components/Snackbar/Snackbar.tsx
narin: Updated the docs

packages/components/src/index.tsx
narin: Moved

packages/components/src/components/Snackbar/Snackbar.stories.tsx
```diff
@@ -34,27 +35,48 @@ export default meta
...
     })
   }
 
-  return <Button label="Press for Snackbar" onPress={onPressSnackbar} />
+  return (
+    <>
+      <Button label="Press for Snackbar" onPress={onPressSnackbar} />
+    </>
+  )
+}
```
TimRoe, line 57: Was this from our weird inconsistency in auto formatting or was it part of draft testing that can be rolled back? At a minimum the `<>` and `</>` seem unneeded since it's only a single component.

packages/components/src/components/Snackbar/Snackbar.stories.tsx
```diff
@@ -34,27 +35,48 @@ export default meta
...
+
+export const _Default: Story = {
```
TimRoe, line 59: Might be worth checking with Jessica if we wanted this named Default or something else as the only one that'll appear in Storybook.

packages/components/src/components/Snackbar/Snackbar.stories.tsx
TimRoe: Fancy 👍

packages/components/src/components/Snackbar/Snackbar.stories.tsx
```diff
@@ -34,27 +35,48 @@ export default meta
 
 type Story = StoryObj<SnackbarProps>
 
+const isWeb = Platform.OS !== 'web'
```
TimRoe, line 38:
```suggestion
const isNotWeb = Platform.OS !== 'web'
```
or flip it to an `===` and switch the conditionals below. 

Not sure how long it has been like that (I think it's new?), but `Platform.OS` also returns 'windows' and 'macos' now so my initial thought of renaming to isMobile would also not technically be accurate. Guess it would if reworked to check if it equals iOS or Android.

TimRoe: Or better yet: delete the variable since now it's not needed for the offset story and only check it on the Default story.

```suggestion
```

packages/components/src/components/Snackbar/SnackbarProvider.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+import { SNACKBAR_DEFAULT_OFFSET, Snackbar } from './Snackbar'
+import { ToastProvider } from 'react-native-toast-notifications'
+
+type SnackbarContextType = {
+  offset: number
+  setOffset: (newOffset: number) => void
+}
+
+export const SnackbarContext = createContext<SnackbarContextType | undefined>(
```
TimRoe, line 11: Not sure if there's special guidance around the create-/use-Context, but should this be named like a global const/variable? AKA should it be all caps w/ underscores to ensure no conflicts with what consumers may be doing?

TimRoe: Nvm, see below it's used as a component so probably good the way it is.

packages/components/src/hooks/useSnackbar.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+import {
+  SNACKBAR_DEFAULT_OFFSET,
+  modifyToastOptions,
+} from '../components/Snackbar/Snackbar'
+import { SnackbarContext } from '../components/Snackbar/SnackbarProvider'
+
+/**
+ * Passthrough of useToast hook which has been modified to allow for optional
+ * offset setting
```
TimRoe, line 13: Think we may want to update this as this is doc that will appear on hover for consumers and they shouldn't need to know or care that we're using a particular library under the hood or how we're modifying it.

Perhaps something like:
```suggestion
 * Hook to manipulate the Snackbar with `show`, `hide`, or `isOpen` to query if already present
 *
 * Must be wrapped in `SnackbarProvider` to function
```
?

packages/components/src/hooks/useSnackbar.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  if (!context) {
+    throw new Error('useSnackbar must be used within a SnackbarProvider')
+  }
+
+  const { offset, setOffset } = context
+
+  const show = (
+    message: string,
+    snackbarOptions?: modifyToastOptions['data'],
```
TimRoe, line 27: Hmmm. With the revisions I wonder if we should change this up now. 

Previously, we were leaning more directly on the library and correspondingly couldn't modify the structure as well. With how this is now, if we wanted to allow modifying the duration in the future we could easily bundle it into the data and would just update the `toast.show` call below to send it in alongside data so it'd be a non-breaking change. 

So, unless I'm mistaken I think we could do some restructuring to simplify a bit/make it more intuitive. Thoughts on splitting out a ticket to rework this a bit? A bit tangential to this ticket.

packages/components/src/components/Snackbar/Snackbar.tsx
TimRoe: Getting around to watching the videos, maybe I am wrong about the `SafeAreaView` part above. I see your iOS device has the bottom bar thing that mine does not (my test device having a physical home button and yours having the bar since it doesn't?). 

Maybe we should leave the default for this ticket and split out another ticket to tackle ensuring we're handling device specific offsets well?

TimRoe:
> You'll notice I separated `useSnackbar` into its own hook file in a new `hooks/` folder. I think our `utils/index.tsx` is getting large enough to point now where we might want to break them up by category, e.g. `hooks/colors.tsx`, `hooks/accessibility.tsx` and have hooks that may be specific to a component be its own file. If we're all onboard with this pattern, I can create a follow-up ticket to split up the rest of our hooks.

I am fine with the splitting things up more, but I'm not sure about the having a dedicated `hooks` folder. I say that because hooks generally are util type functions and we also have non-hook utils. Seems like it might create a weird separation of generally similar functions. 

For example:
- `storybook.tsx` has `webStorybookColorScheme` which is a hook and `generateDocs` which really isn't despite naming with lower case
- `OSfunctions.tsx` similarly has a hook and non-hook util (two if counting the global `isIOS` although that's used kinda like a hook)

Just seems like it could leave us in an area of fairly arbitrary delineation of similarly scoped utils some in the `hooks` and some in `utils`. 

Also, in terms of this case specifically, I wonder if the useSnackbar hook file shouldn't just be in the component folder still? I somewhat started the precedent with the icon script putting the `iconList.ts` file in with the component despite it not being strictly the component/tests/stories file--it is intrinsically tied to the component.

TimRoe (changes requested):
Looks mostly good and like how it came together, but dropping a request changes to denote finishing a first look through.

Some minor comments suggesting further quick fixes, the more notable outstanding comments being:
- [Default offset](https://github.com/department-of-veterans-affairs/va-mobile-library/pull/423#discussion_r1707770736) - maybe just split to a separate ticket
- Maybe [simplify data typing](https://github.com/department-of-veterans-affairs/va-mobile-library/pull/423#discussion_r1707934581) - make separate ticket if my thoughts make sense to you as well
- [Thought on the hooks folder](https://github.com/department-of-veterans-affairs/va-mobile-library/pull/423#issuecomment-2274419136)

packages/components/src/components/Snackbar/Snackbar.stories.tsx
```diff
@@ -34,27 +35,48 @@ export default meta
...
+    data: {
+      isError: false,
+      messageA11y: 'Message moved to Custom Folder with a11y',
+      onActionPressed: () => console.log('Action pressed'),
+    },
+  },
 }
 
-export const _Snackbar: Story = {
-  render: Platform.OS !== 'web' ? mobileComponentRenderer : undefined, // Render Snackbar flat in web
```
TimRoe, line 51: CustomOffset story ended up losing the render and just renders flat--still needs:
```
render: mobileComponentRenderer,
```

packages/components/src/components/Snackbar/Snackbar.stories.tsx
narin: Ended up removing the Custom Offset story as I couldn't get conditional naming of the story to work. I figure it'd be pretty easy for the user to add offset in the args

packages/components/src/hooks/useSnackbar.tsx
narin: Created ticket #439

narin:
- Addressed all nits
- Deleted `hooks/` folder and went with suggestion that component specific hook should probably live within the component folder itself
- Ended up removing Custom Offset story since there seems to be a bug with setting story names conditionally for mobile/web.
- Updated `useSnackbar` comment to remove reference to useToast
- Set default `offset` based on OS to match app. Follow-up ticket to look into how it plays with react-native-safe-area-context here: #438 
- Created new ticket to simplify types: #439

packages/components/src/components/Snackbar/Snackbar.stories.tsx
```diff
@@ -5,7 +5,7 @@ import React from 'react'
 import { Button } from '../Button/Button'
 import { Snackbar, SnackbarProps } from './Snackbar'
 import { generateDocs } from '../../utils/storybook'
-import { useSnackbar } from '../../hooks/useSnackbar'
+import { useSnackbar } from '../Snackbar/useSnackbar'
```
TimRoe, line 8:
```suggestion
import { useSnackbar } from './useSnackbar'
```

packages/components/src/components/Snackbar/Snackbar.tsx
```diff
@@ -18,7 +19,7 @@ import { Spacer } from '../Spacer/Spacer'
 import { useTheme } from '../../utils'
 
 // TODO: Replace with global setting
-export const SNACKBAR_DEFAULT_OFFSET: number = 50
+export const SNACKBAR_DEFAULT_OFFSET: number = Platform.OS === 'ios' ? 25 : 0
```
TimRoe, line 22: Will probably change either on the offset follow-up story or the unit tests one, but just noting we have the isIOS util function as a dedicated shorthand.

packages/components/.storybook/web/manager.ts
```diff
@@ -0,0 +1,11 @@
+import { addons } from '@storybook/manager-api'
+
+addons.setConfig({
```
TimRoe, line 3: Not sure if we want to clean this up now that it's no longer used. Seems handy to have around if we need it, but not sure if we'll ever use it again.

TimRoe (approved):
Approved with a couple additional nitpicky suggestions.

Issue #370, opened 2024-06-26
DS - SnackBar Component Flagship Work Setup


kellylein:
## Description 
This is a ticket for Product to create an epic and work with Flagship to plan the work.

### Notes
- 

## Acceptance Criteria
  

-  [ ]   Discuss with the team any specifics of implementing this component into the app
-  [x]   A ticket has been created for the app teams to install the new component  -  #9560, #9561

TimRoe:
For this AC:
> Discuss with the team any specifics of implementing this component into the app

some technical recommendations from tech discovery (may change during implementation):
- Remove `bottomOffset` from redux and generally remove the SnackBar (app level name, design system is Snackbar) component slice
   - Design system Snackbar component will handle setting when showing Snackbar (since it is dismissed on navigation, should not need updating height)
- Should be able to clean up `global.d.ts` and `src/App.tsx` of custom config alongside plugging in the design system Snackbar per documentation
- While calling the Snackbar directly via `snackbar.show(...)` and `snackbar.hideAll()` will be possible, it is recommended to use design system provided convenience functions when possible as they bake in logic and automatically adjust to changes where manual calls may need more updating
- Remove `SnackBarConstants` from `src/constants/common.ts` as no longer relevant
- Recommend removal of `SnackbarMessages` typing that mandates success-/error-Msg properties and conditionally includes undo(Error)Msg
   - Depends on screen, but rarely seems to be beneficial vs just plugging in the translation text--context both from surrounding logic and translation name clarifies what the Snackbar is appearing for
- Should be able to clean up appreciable snackbar-related parts of app-level theme

TimRoe:
Additional note:
- When testing announcing the Snackbar for accessibility, some conflicting behavior with app-level setting focus to the back button was observed as documented on [this PR](https://github.com/department-of-veterans-affairs/va-mobile-library/pull/437)
   - Unclear how this should be resolved, but something to be cognizant during implementation into flagship

kellylein:
Snackbar accessibility functions correctly with RN 0.74 or higher.  Flagship is currently on 0.73.

PR #437, opened 2024-08-09
[Feature] Announce Snackbar appearance to screen reader

TimRoe:
## Description of Change

Recognized the issue with the screen reader not reading the `announceForAccessibility` was only a problem in iOS and it appeared to be an issue due to the synchronous nature of the storybook test case so iOS was immediately refocusing the button at near enough the same time to not do the `announceForAccessibility`. 

Resolved by creating a useEffect on the Snackbar display that:
1. Has an empty dependency array so the useEffect is only run once on initial render since otherwise the `announceForAccessibility` was being triggered multiple times (on appearance and disappearance)
2. Has a 50ms delay so that if whatever is bringing up the Snackbar is synchronous (not generally how it's used, but could be) it still announces correctly

#### Testing Packages

- [0.21.1-alpha.10](https://www.npmjs.com/package/@department-of-veterans-affairs/mobile-component-library/v/0.21.1-alpha.10)

### Screenshots/Video

iOS:

https://github.com/user-attachments/assets/335e4726-ab05-431f-909b-fdffc6ccaaea

Android:

https://github.com/user-attachments/assets/5f889ace-fd70-45d7-bf2c-9c1e185d4f37

Flagship iOS (somewhat janky from app-level focus that they'll need to address during implementation):

https://github.com/user-attachments/assets/571f55bc-cd8a-4a15-a881-8123cdda436e

Flagship Android (also somewhat janky in a different way from app-level focus that they'll need to address during implementation):

https://github.com/user-attachments/assets/046f41f4-a28e-433f-bb4d-632be14b0641

## Testing

Tested it performs as expected in Storybook. Also tested flagship which was a bit janky, but due to app-level setting focus.

- [x] Tested on iOS 
- [x] Tested on Android 
- Tested on Web 
   - N/A

## PR Checklist
Code reviewer validation:
- General
	- [x] Acceptance criteria: 
		- All satisfied _or_
- Code

## Publish

While changes do warrant a new version [per the versioning guidelines](https://github.com/department-of-veterans-affairs/va-mobile-library#versioning-policy), postponed for later following functional changes to Snackbar.


narin (approved):
Looks good. Did you identify how the app would fix the jankiness? If so we should make note of it in the implementation ticket.

TimRoe:
> Did you identify how the app would fix the jankiness?

I did not. 

It may involve changes on our end. There is a slightly different function, [announceForAccessibilityWithOptions](https://reactnative.dev/docs/accessibilityinfo#announceforaccessibilitywithoptions), which can make iOS queue up the announcement behind other screen reader behavior similar the how Android behaves. There's also settings for Android to be assertive and interrupt, but not sure how easily we can apply that to `announceForAccessibility`. There's enough ambiguity that I think we can leave it to flagship implementing and work with them on how best to sort it out then.

I have added a [comment to the handoff ticket](https://github.com/department-of-veterans-affairs/va-mobile-library/issues/370#issuecomment-2284643658) so they are on the lookout for it at least.

Issue #440, opened 2024-08-12
DS - Snackbar - Improve screen reader behavior


TimRoe:
## Description 
There was [a slack thread discussion](https://adhoc.slack.com/archives/C05DQPHUVNZ/p1723231477940539) on whether the behavior called for by the Figma doc for Snackbar:
> "Snackbar or toast must be dynamically announced by the screen reader without moving focus to it."
Current: App automatically moves the focus to the snackbar.
Proposed: Snackbar is announced, but focus is not moved.

could be further improved for blind/low-vision (screen reader) users. For example: auto-dismissing it when there's no action to take and it's purely informational.

## Acceptance Criteria
  

-  [ ] Determine if further behavior refinement is desirable (e.g. auto-dismissing the Snackbar after 5 seconds if there's no follow-up action to Undo/Try again) and, if so, implement it

**Accessibility Requirements**
- [ ] Behavior enhanced according to specifications worked out with Brea

PR #442, opened 2024-08-13
[Feature] Snackbar – Update default offset to assume nav bar

narin:
## Description of Change

Updates default offset to match app and assume there is a nav bar present. Values copied from [va-mobile-app's standardTheme.ts](https://github.com/department-of-veterans-affairs/va-mobile-app/blob/3fd975d714b3ed4628353b343270f1d0c6291c3c/VAMobile/src/styles/themes/standardTheme.ts#L301).

Also adds util to check for iOS devices with home buttons and adjust offset depending on that. 

Default offset will eventually be replaced by a global setting that can be adjusted.

In regards to the safe areas, based on my research, nesting multiple <SafeAreaProvider> or using <SafeAreaView> within <SafeAreaProvider>s is not recommended. The mobile app already has a SafeAreaProvider so we won't include it here but made note of it in the docs.

#### Testing Packages

- 0.21.1-alpha.11
- 0.21.1-alpha.12
- 0.21.1-alpha.13
- 0.21.1-alpha.15

### Screenshots/Video

<details><summary>DS snackbar vs App Snackbar (iOS)</summary>
</details>

<details><summary>DS snackbar vs App Snackbar (iPhone SE with home button)</summary>
</details>

<details><summary>DS snackbar vs App Snackbar (Android)</summary></details>

## Testing

- [x] Tested on iOS 
- [x] Tested on Android 
- [x] Tested on Web

## PR Checklist
Code reviewer validation:
- General
	- [x] Acceptance criteria: 
		- All satisfied _or_
- Code

## Publish

Doesn't warrant publish.  Will publish after #440 since that should be the last ticket with functional changes.

### DIFF
--- a/packages/components/src/components/Snackbar/Snackbar.tsx
+++ b/packages/components/src/components/Snackbar/Snackbar.tsx
@@ -49,29 +49,6 @@ export type SnackbarType = Omit<ToastType, 'show' | 'update'> & {
   ) => string
 }
 
-//
-// TODO: May need to move the following doc to the custom render to show in Storybook
-//
-
-/**
- * To use SnackbarProvider, your app must have a global.d.ts which you can copy the following into:
- * ```
- * type SnackbarType = import('@department-of-veterans-affairs/mobile-component-library').SnackbarType
- * // eslint-disable-next-line no-var
- * declare var snackbar: SnackbarType
- * ```
- * then add SnackbarProvider in App.tsx (or similar foundational file) level with your app rendering:
- * ```
- * <>
- *   <App />
- *   <SnackbarProvider />
- * </>
- * ```
- *
- * This config will allow it to be called anywhere including outside React components.
- *
- * The Snackbar remains open indefinitely. App configuration should ensure it is dismissed on navigation.
- */
 export const SnackbarProvider: React.FC = () => {
   // const forceUpdate = useReducer(x => x + 1, 0)[1]
   // const [offset, setOffset] = useState(10)
@@ -136,6 +113,25 @@ const SnackbarButton: FC<SnackbarButtonProps> = ({ text, onPress }) => {
 
 export type SnackbarProps = Omit<ToastProps, 'data'> & snackbarData
 
+/**
+ * To use SnackbarProvider, your app must have a global.d.ts which you can copy the following into:
+ * ```
+ * type SnackbarType = import('@department-of-veterans-affairs/mobile-component-library').SnackbarType
+ * // eslint-disable-next-line no-var
+ * declare var snackbar: SnackbarType
+ * ```
+ * then add SnackbarProvider in App.tsx (or similar foundational file) level with your app rendering:
+ * ```
+ * <>
+ *   <App />
+ *   <SnackbarProvider />
+ * </>
+ * ```
+ *
+ * This config will allow it to be called anywhere including outside React components.
+ *
+ * The Snackbar remains open indefinitely. App configuration should ensure it is dismissed on navigation.
+ */
 export const Snackbar: FC<SnackbarProps> = (toast) => {
   const fontScale = useWindowDimensions().fontScale
   const theme = useTheme()
@@ -221,7 +217,7 @@ export const Snackbar: FC<SnackbarProps> = (toast) => {
 
   const buttonContainer: ViewProps = {
     style: {
-      flex: 0,
+      flexGrow: 0,
       flexDirection: 'row',
       flexWrap: 'wrap',
       marginLeft: 'auto', // Maintains alignment to right side
--- a/packages/components/src/components/Snackbar/Snackbar.tsx
+++ b/packages/components/src/components/Snackbar/Snackbar.tsx
@@ -82,17 +82,17 @@ export type SnackbarProps = Omit<ToastProps, 'data' | 'message'> &
  *
  * ```jsx
  * return (
- *   <SnackbarProvider>
- *     <App />
- *   </SnackbarProvider>
+ *   <SafeAreaProvider
+ *     <SnackbarProvider>
+ *       <App />
+ *     </SnackbarProvider>
+ *   </SafeAreaProvider
  * )
  * ```
  *
- * **Note:** If you are using safe area helpers such as [SafeAreaView](https://reactnative.dev/docs/safeareaview)
- * from React Native or SafeAreaProvider
- * from [react-native-safe-area-context](https://github.com/th3rdwave/react-native-safe-area-context)
- * we recommend you place the SnackbarProvider within these safe area views, otherwise default
- * offset from the bottom of the screen may not display as expected.
+ * **Note:** The snackbar requires [react-native-safe-area-context](https://github.com/th3rdwave/react-native-safe-area-context).
+ * SnackbarProvider should be placed within SafeAreaProvider. If you do not have a SafeAreaProvider, you can use
+ * SnackbarWithSafeAreaProvider instead.
  *
  * Then within any component, import the `useSnackbar` hook and use the .show() or .hide()
  * methods to display a Snackbar:
--- a/packages/components/src/components/Snackbar/Snackbar.tsx
+++ b/packages/components/src/components/Snackbar/Snackbar.tsx
@@ -1,4 +1,5 @@
 import {
+  AccessibilityInfo,
   Pressable,
   PressableStateCallbackType,
   Text,
@@ -11,7 +12,7 @@ import {
 } from 'react-native'
 import { ToastProps } from 'react-native-toast-notifications/lib/typescript/toast'
 import { useTranslation } from 'react-i18next'
-import React, { FC } from 'react'
+import React, { FC, useEffect } from 'react'
 
 import { Icon, IconProps } from '../Icon/Icon'
 import { Spacer } from '../Spacer/Spacer'
@@ -74,7 +75,10 @@ const SnackbarButton: FC<SnackbarButtonProps> = ({ text, onPress }) => {
   )
 }
 
-export type SnackbarProps = Omit<ToastProps, 'data'> & snackbarData
+export type SnackbarProps = Omit<ToastProps, 'data' | 'message'> &
+  snackbarData & {
+    message: string
+  }
 
 /**
  * To use SnackbarProvider, import SnackbarProvider in App.tsx (or similar foundational file) and
@@ -106,10 +110,18 @@ export const Snackbar: FC<SnackbarProps> = (toast) => {
   const fontScale = useWindowDimensions().fontScale
   const theme = useTheme()
   const { t } = useTranslation()
-  // const a11yAnnounce =
-  //   AccessibilityInfo.announceForAccessibility('Test announce two')
 
-  // useEffect(() => a11yAnnounce, [a11yAnnounce])
+  /**
+   * useEffect to handle announcing the Snackbar appearing to the screen reader
+   */
+  useEffect(() => {
+    // Delay to prevent iOS from instantly refocusing the action prompting the Snackbar if synchronous
+    setTimeout(
+      () => AccessibilityInfo.announceForAccessibility(messageA11y || toast.message),
+      50,
+    )
+    // Empty dependency array so useEffect only runs on initial render
+  }, [])
 
   const helperText: TextStyle = {
     fontFamily: 'SourceSansPro-Regular',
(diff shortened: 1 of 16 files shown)


packages/components/src/utils/OSfunctions.tsx
```diff
@@ -110,3 +110,29 @@ export function useExternalLink(): (
...
   }
 }
+
+/**
+ * Returns whether or not the device has a home button
+ */
+export const hasHomeButton = () => {
+  const { height, width } = Dimensions.get('window')
+
+  if (Platform.OS !== 'ios') {
```
TimRoe, line 120:
```suggestion
  if (!isIOS()) {
```

TimRoe: Nitpicky: could also move above the Dimensions call since it just gets dropped on the floor

packages/components/src/utils/OSfunctions.tsx
```diff
@@ -110,3 +110,29 @@ export function useExternalLink(): (
...
+    { height: 568, width: 320 }, // iPhone 5, 5s, 5c, SE (1st gen)
+  ]
+
+  const currentDevice = { height, width }
+
+  return devicesWithHomeButton.some(
+    (device) =>
+      device.height === currentDevice.height &&
+      device.width === currentDevice.width,
+  )
```
TimRoe, line 137: I'm not sure about this. It feels brittle like new devices (e.g. iPhone SE 4th gen) could have a new resolution and things could break on us--especially since the list doesn't note the SE (3rd gen), but looks like it has the same resolution as the 2nd gen.

Generally: not sure how I feel about us being in the business of manually handling this.

narin: Do you have an alternative suggestion? iPhone SE 4th gen [is rumored](https://www.macrumors.com/guide/iphone-se-4/) to no longer have a home button so this list shouldn't change.

TimRoe: We couldn't do native module stuff at all without converting the project, right? Had found [this](https://stackoverflow.com/questions/52652956/how-to-detect-if-the-device-iphone-has-physical-home-button) on basically querying the insets without directly using them.

Otherwise I guess not if not leaning on `react-native-safe-area-context` in some capacity, whether directly or using as a means to do a similar query to the above without native.

packages/components/src/utils/OSfunctions.tsx
```diff
@@ -110,3 +110,29 @@ export function useExternalLink(): (
     }
   }
 }
+
+/**
+ * Returns whether or not the device has a home button
+ */
+export const hasHomeButton = () => {
```
TimRoe, line 117: Should be capitalized?

TimRoe (approved):
Approved with some outstanding nitpicky things. Think the only notable think is a separate ticket to add unit tests  for the new function so we don't forget.

narin:
Since we know the app is using `react-native-safe-area-context` decided to leverage that instead and provider a helper provider for apps not using it.
- Created a new hook `useSnackbarDefaultOffset()` which uses `useSafeAreaInsets()` bottom inset to determine default offset
- Fixed Android offset so it's the same as iOS
- Add `<SnackbarWithSafeArea/>` convenience provider for apps that don't have a `<SafeAreaProvider/>`
- Remove `hasHomeButton()` util

I assume by the time another app is using our lib, we will have figured out native code by then and could switch to using `react-native-device-info` or similar libs to check for home buttons.

packages/components/src/components/Snackbar/Snackbar.tsx
```diff
@@ -90,17 +82,17 @@ export type SnackbarProps = Omit<ToastProps, 'data' | 'message'> &
...
  * ```
  *
- * **Note:** If you are using safe area helpers such as [SafeAreaView](https://reactnative.dev/docs/safeareaview)
- * from React Native or SafeAreaProvider
- * from [react-native-safe-area-context](https://github.com/th3rdwave/react-native-safe-area-context)
- * we recommend you place the SnackbarProvider within these safe area views, otherwise default
- * offset from the bottom of the screen may not display as expected.
+ * **Note:** The snackbar requires [react-native-safe-area-context](https://github.com/th3rdwave/react-native-safe-area-context).
+ * SnackbarProvider should be placed within SafeAreaProvider. If you do not have a SafeAreaProvider, you can use
+ * SnackbarWithSafeAreaProvider instead.
```
TimRoe, line 95:
```suggestion
 * SnackbarProviderWithSafeArea instead.
```

packages/components/src/components/Snackbar/SnackbarProvider.tsx
```diff
@@ -1,7 +1,9 @@
 import React, { ReactNode, createContext, useState } from 'react'
 
-import { SNACKBAR_DEFAULT_OFFSET, Snackbar } from './Snackbar'
+import { SafeAreaProvider } from 'react-native-safe-area-context'
+import { Snackbar, SnackbarProps } from './Snackbar'
 import { ToastProvider } from 'react-native-toast-notifications'
```
TimRoe, line 5: Nitpicky: mixing library and internal file imports

packages/components/src/components/Snackbar/useSnackbarDefaultOffset.tsx
```diff
@@ -0,0 +1,17 @@
...
+  const insets = useSafeAreaInsets()
+
+  if (!insets) {
+    throw new Error(
+      'useSnackbarDefaultOffset must be used within a SafeAreaProvider. Use SnackbarProviderWithSafeArea or add a SafeAreaProvider with react-native-safe-area-context.',
+    )
+  }
+
+  const { bottom } = insets
+  return bottom === 0 ? 60 : 60 + bottom
```
TimRoe, line 16:
```suggestion
  return 60 + insets.bottom
```

TimRoe: May also be worth making 60 a const or having a comment noting it represents the height of the Nav bar

packages/components/src/components/Snackbar/Snackbar.tsx
```diff
@@ -86,12 +82,18 @@ export type SnackbarProps = Omit<ToastProps, 'data' | 'message'> &
...
- *   </SnackbarProvider>
+ *   <SafeAreaProvider
+ *     <SnackbarProvider>
+ *       <App />
+ *     </SnackbarProvider>
+ *   </SafeAreaProvider
  * )
  * ```
  *
+ * **Note:** The snackbar requires [react-native-safe-area-context](https://github.com/th3rdwave/react-native-safe-area-context).
```
TimRoe, line 93: Need to move `react-native-safe-area-context` from a devDependency to dependency for it to work in consuming apps that wouldn't otherwise have it.

TimRoe (changes requested):
Good revisions, makes much more sense in my mind.

Marking request changes because we need to change `react-native-safe-area-context` to a dependency or `SnackbarProviderWithSafeArea` wouldn't actually work within a consuming app without them also adding the package.

Beyond that:
- Some minor nitpick suggestions here and there
- Don't feel strongly, but wonder if `useSnackbar` and `useSnackbarDefaultOffset` instead of being their own files should be together in a `hooks.tsx` or something (still within the Snackbar folder)

packages/components/src/components/Snackbar/Snackbar.tsx
```diff
@@ -165,6 +172,10 @@ export const Snackbar: FC<SnackbarProps> = (toast) => {
     },
   }
 
+  if (isAndroid()) {
+    containerProps.tabIndex = 0
+  }
+
```
narin, line 178: Had to move this out because in the app, running on iOS would get the following error:

`node_modules/@department-of-veterans-affairs/mobile-component-library/src/components/Snackbar/Snackbar.tsx(160,5): error TS2353: Object literal may only specify known properties, and 'tabIndex' does not exist in type 'ViewProps'.`

narin:
- Addressed nits
- Putting `react-native-safe-context` as a dependency caused a conflict and crash within the app. Put it in `peerDependencies` instead
- Added `isAndroid()` check to the `tabIndex` for the Snackbar container props to avoid TS error when running in app on iOS
- Bumped mobile-tokens in peerDependencies since the new color token from 0.14.0 is needed in Snackbar
- Keeping hooks separated. I think I tried this initially but was getting a circular dependency warning

packages/components/src/components/Snackbar/useSnackbar.tsx
```diff
@@ -20,11 +20,9 @@ export function useSnackbar() {
...
   }
 
-  const show = (
-    message: string,
-    snackbarOptions?: modifyToastOptions['data'],
-  ) => {
-    const { offset, setOffset } = context
+  const { offset, setOffset } = context
+
+  const show = (message: string, snackbarOptions?: SnackbarOptions) => {
```
TimRoe, line 25: Doesn't matter, but just noting that `const { offset, setOffset } = context` was moved out of `show`. Not sure if intentional or just a minor inconsistency fixing conflicts with my changes.

packages/components/src/components/Snackbar/Snackbar.tsx
TimRoe: Weird. I had noticed TS giving a red squiggly line when prototyping in the app node_modules on one ticket, but it seemed to run fine in iOS/Android previously. Oh well.

packages/components/src/components/Snackbar/Snackbar.tsx
```diff
@@ -181,7 +158,6 @@ export const Snackbar: FC<SnackbarProps> = (toast) => {
 
   const containerProps: ViewProps = {
     accessibilityViewIsModal: true, // iOS only
-    tabIndex: 0, // Android only
     // Above props prevent screen reader from tap focusing elements behind the Snackbar
```
TimRoe, line 161: Think we need to duplicate this comment or otherwise rework slightly for clarity--maybe pull the iOS one out as well for something like:
```
  // Below props prevent screen reader from tap focusing elements behind the Snackbar
  if (isAndroid()) {
    containerProps.tabIndex = 0
  }
  if (isIOS()) {
    containerProps.accessibilityViewIsModal = true
  }
```
?

packages/components/src/components/Snackbar/SnackbarProvider.tsx
```diff
@@ -1,7 +1,9 @@
+import { ToastProvider } from 'react-native-toast-notifications'
 import React, { ReactNode, createContext, useState } from 'react'
```
TimRoe, line 2:
```suggestion
import { SafeAreaProvider } from 'react-native-safe-area-context'
import { ToastProvider } from 'react-native-toast-notifications'
import React, { ReactNode, createContext, useState } from 'react'
```

packages/components/src/components/Snackbar/SnackbarProvider.tsx
```diff
@@ -1,7 +1,9 @@
...
 
-import { SNACKBAR_DEFAULT_OFFSET, Snackbar, SnackbarProps } from './Snackbar'
-import { ToastProvider } from 'react-native-toast-notifications'
+import { SafeAreaProvider } from 'react-native-safe-area-context'
```
TimRoe, line 4:
```suggestion

```
with above suggestion

packages/components/src/components/Snackbar/SnackbarProvider.tsx
```diff
@@ -24,10 +28,18 @@ export const SnackbarProvider: React.FC<{ children: ReactNode }> = ({
         duration={1000000000000} // Essentially indefinite until dismissed
         offset={offset}
         placement="bottom"
-        renderToast={(toast) => <Snackbar {...toast as SnackbarProps} />}
+        renderToast={(toast) => <Snackbar {...(toast as SnackbarProps)} />}
```
TimRoe, line 31:
```suggestion
        renderToast={(toast) => <Snackbar {...toast as SnackbarProps} />}
```
?

TimRoe (approved):
Approved w/ some final minor nitpicks.

packages/components/src/components/Snackbar/SnackbarProvider.tsx
narin: Autoformatter forces this for some reason. I can save without formatting

packages/components/src/components/Snackbar/useSnackbar.tsx
narin: Will be cleaned up further in the screenreader dismiss ticket #440

PR #443, opened 2024-08-13
[Feature] Snackbar - Simplify Typing

TimRoe:
## Description of Change

Reworked the Snackbar typing to be more clear.

Unrelated to ticket: added `prettier-eslint` package to base repo to resolve the Prettier ESLint plugin from getting an error when running instead of simultaneously handling both Prettier and ESLint formatting issues.

## Testing

Spot checked each platform to confirm nothing was broken by type restructuring.

- [x] Tested on iOS 
- [x] Tested on Android 
- [x] Tested on Web

## PR Checklist
Code reviewer validation:
- General
	- [x] Acceptance criteria: 
		- All satisfied _or_
- Code

## Publish

While changes warrant a new version [per the versioning guidelines](https://github.com/department-of-veterans-affairs/va-mobile-library#versioning-policy), we are postponing Snackbar publish until later.

### DIFF
--- a/packages/components/src/components/Snackbar/Snackbar.tsx
+++ b/packages/components/src/components/Snackbar/Snackbar.tsx
@@ -120,12 +120,10 @@ export const Snackbar: FC<SnackbarProps> = (toast) => {
    * useEffect to handle announcing the Snackbar appearing to the screen reader
    */
   useEffect(() => {
+    const announcement = messageA11y || toast.message
     // Delay to prevent iOS from instantly refocusing the action prompting the Snackbar if synchronous
     setTimeout(
-      () =>
-        AccessibilityInfo.announceForAccessibility(
-          messageA11y || toast.message,
-        ),
+      () => AccessibilityInfo.announceForAccessibility(announcement),
       50,
     )
     // Empty dependency array so useEffect only runs on initial render
--- a/packages/components/src/components/Snackbar/Snackbar.tsx
+++ b/packages/components/src/components/Snackbar/Snackbar.tsx
@@ -22,27 +22,6 @@ import { useTheme } from '../../utils'
 // TODO: Replace with global setting
 export const SNACKBAR_DEFAULT_OFFSET: number = isIOS() ? 25 : 0
 
-type snackbarData = {
-  data?: {
-    /** true if snackbar represents an error state */
-    isError?: boolean
-    /** message text A11y label override */
-    messageA11y?: string
-    /** offset from bottom of screen. defaults to 50 in SnackbarProvider */
-    offset?: number
-    /** action button onPress logic for "Try again" (isError=true) or "Undo" button */
-    onActionPressed?: () => void
-  }
-}
-
-/**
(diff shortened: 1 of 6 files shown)


packages/components/src/components/Snackbar/Snackbar.tsx
```diff
@@ -75,10 +54,36 @@ const SnackbarButton: FC<SnackbarButtonProps> = ({ text, onPress }) => {
...
+// List of ToastProps needed within Snackbar
+type activeToastProps = 'onHide'
+
+/**
+ * Properties used by Snackbar component
+ */
+export type SnackbarProps = Pick<ToastProps, activeToastProps> & {
+  data?: SnackbarData
+  message: string
+}
```
TimRoe, line 86: This is the part mentioned in standup that I'm not sure is worth it.

On the one hand, this nicely delineates the typing into purposeful bins:
- Options - what end users can send to `useSnackbar.show`
- Data - what goes into the actually rendered Snackbar (excluding toast-level thing like offset, potentially duration)
- Props - what's needed in the rendered Snackbar

On the other hand, it adds complexity that I'm not sure is worth it vs:
- Merge SnackbarData into SnackbarOptions and:
```
export type SnackbarProps = Omit<ToastProps, 'data' | 'message'> &
  {
    data?: SnackbarOptions
    message: string
  }
```
and having a bunch of extra irrelevant stuff (both library-level and our props that are irrelevant in the component itself) in the Snackbar when using `toast.`

narin: I'm in favor of the solution that locks it down more which sounds like its the more complex one. 

Is there a functional reason why `offset` isn't just included in `SnackbarData`, or is it solely to separate them by purpose? That could be a middle ground that is still relatively clean.

narin (approved):
Looks good. I'm in favor of a solution that locks down the props to avoid manipulation outside of the intended design. One suggestion to clean up typings if applicable.

packages/components/src/components/Snackbar/Snackbar.tsx
TimRoe: `offset` is separate because it's not relevant to pass into the Snackbar itself since it's only relevant to the hook/SnackbarProvider. So in the Snackbar you'd be able to reference `toast.offset` which is meaningless by that point.

In terms of what the consumer sees, this typing is no different than that simplified example. 

It is purely for our benefit and a fairly minor one at that--basically just controlling what appears in Snackbar when you do `toast.`. So really the benefit question is about if having it structured helps make it clearer what each piece is for or if we don't really care that a bunch of needless props appear for `toast.` and want to keep it simple.

narin: Okay, I understand now. Thinking long term for engineers that join or if we circle back, if we're pretty sure those additional props from `toast.` aren't needed, I think it could be useful to hide them so they don't show up in Intellisense.

Issue #447, opened 2024-08-19
DS - Add Spacing Tokens to Token Package


kellylein:
## Description 
As part of this story we want to:   
- Add the spacing tokens into the tokens package so they can be used. There are 12 spacing tokens to add.  They are currently in Figma, but Jessica will look at exporting them as JSON.

## Design Notes
- [Notes](https://www.figma.com/design/XQwpzcNSGyGJN4nXDPoEj0/Design-Tokens---%F0%9F%94%8D-Design-System---VA-Mobile?node-id=225-1140&t=vQm9WlaU68BEyqVD-4)
- [JSON export from Figma](https://www.figma.com/design/XQwpzcNSGyGJN4nXDPoEj0/Design-Tokens---%F0%9F%94%8D-Design-System---VA-Mobile?node-id=224-11312&t=vQm9WlaU68BEyqVD-4)
- [Spacing tokens](https://www.figma.com/design/XQwpzcNSGyGJN4nXDPoEj0/Design-Tokens---%F0%9F%94%8D-Design-System---VA-Mobile?node-id=224-11338&t=vQm9WlaU68BEyqVD-4)

<img width="1081" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd0t6Qmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--f0414bbb79291a29a76d1acae0b2b62c2efebb00/Screenshot%202024-09-05%20at%2010.23.20%E2%80%AFAM.png" alt="Screenshot 2024-09-05 at 10.23.20 AM.png" />

## Acceptance Criteria
  

- [ ] Determine JSON structure and new file for spacing tokens will live in
- [ ] Add 12 new spacing tokens are in the tokens package
- [ ] Update style-dictionary config to handle spacing tokens and export to a .ts file in our `/dist` folder

Open Questions

- Dependencies/Roadblocks:
    - Any internal/external dependencies?
    - Test accounts needed?
- Does this require QA?
    
- Dev Notes: 

## Ticket Checklist

- [x] Acceptance criteria defined
- [x] Labels added (front-end, back-end, feature)
- [x] Linked to an Epic

PR #474, opened 2024-09-10
[Feature] Update Spacer component to use spacing tokens

narin:
## Description of Change

- Updates Spacer component to use new spacing tokens from #447 
- Default size is now vadsSpaceSm (12)
  - **Note:** Alert was the only component using Spacer with the default value of 10. Confirmed that it is to be updated to vadsSpaceSm (12) as per [Figma doc](https://www.figma.com/design/XQwpzcNSGyGJN4nXDPoEj0/Design-Tokens---%F0%9F%94%8D-Design-System---VA-Mobile?node-id=238-927&node-type=instance&t=UiWJGxavr7qams0k-0) and spacing above buttons in expandable alert is to be updated to vadsSpace4xl (36) vs the old 34
- Updated any components using Spacer with values passed to use new size variants

#### Testing Packages

- components@0.23.1-alpha.1

#### Before/after: 
&nbsp;&nbsp;

## Testing

- [ ] Tested on iOS 
- [ ] Tested on Android 
- [ ] Tested on Web

## PR Checklist
Code reviewer validation:
- General
	- [x] Acceptance criteria: 
		- All satisfied _or_
- Code

## Publish

- [ ] Merge branch to `main`

### DIFF
--- a/packages/components/src/components/Spacer/Spacer.test.tsx
+++ b/packages/components/src/components/Spacer/Spacer.test.tsx
@@ -8,12 +8,12 @@ describe('Spacer', () => {
 
     const spacer = screen.root
 
-    expect(spacer.props.style.height).toBe(10)
+    expect(spacer.props.style.height).toBe(12)
     expect(spacer.props.style.width).toBe('auto')
   })
 
   it('renders correctly at overridden size 20', () => {
-    render(<Spacer size={20} />)
+    render(<Spacer size="lg" />)
 
     const spacer = screen.root
 
@@ -27,11 +27,11 @@ describe('Spacer', () => {
     const spacer = screen.root
 
     expect(spacer.props.style.height).toBe('auto')
-    expect(spacer.props.style.width).toBe(10)
+    expect(spacer.props.style.width).toBe(12)
   })
 
   it('renders horizontally at overridden size 20', () => {
-    render(<Spacer size={20} horizontal />)
+    render(<Spacer size="lg" horizontal />)
 
     const spacer = screen.root
 
--- a/packages/components/src/components/Spacer/Spacer.test.tsx
+++ b/packages/components/src/components/Spacer/Spacer.test.tsx
@@ -3,7 +3,7 @@ import { render, screen } from '@testing-library/react-native'
 import React from 'react'
 
 describe('Spacer', () => {
-  it('renders correctly at default size (10)', () => {
+  it('renders correctly at default size vadsSpaceSm', () => {
     render(<Spacer />)
 
     const spacer = screen.root
@@ -21,7 +21,7 @@ describe('Spacer', () => {
     expect(spacer.props.style.width).toBe('auto')
   })
 
-  it('renders horizontally at default size (10)', () => {
+  it('renders horizontally at default size vadsSpaceSm', () => {
     render(<Spacer horizontal />)
 
     const spacer = screen.root
--- a/packages/components/src/components/Spacer/Spacer.test.tsx
+++ b/packages/components/src/components/Spacer/Spacer.test.tsx
@@ -3,7 +3,7 @@ import { render, screen } from '@testing-library/react-native'
 import React from 'react'
 
 describe('Spacer', () => {
-  it('renders correctly at default size vadsSpaceSm', () => {
+  it('renders correctly at default size  vadsSpaceSm (12)', () => {
     render(<Spacer />)
 
     const spacer = screen.root
@@ -12,7 +12,7 @@ describe('Spacer', () => {
     expect(spacer.props.style.width).toBe('auto')
   })
 
-  it('renders correctly at overridden size vadsSpaceLg', () => {
+  it('renders correctly at overridden size vadsSpaceLg (20)', () => {
     render(<Spacer size="lg" />)
 
     const spacer = screen.root
@@ -21,7 +21,7 @@ describe('Spacer', () => {
     expect(spacer.props.style.width).toBe('auto')
   })
 
-  it('renders horizontally at default size vadsSpaceSm', () => {
+  it('renders horizontally at default size  vadsSpaceSm (12)', () => {
     render(<Spacer horizontal />)
 
     const spacer = screen.root
@@ -30,7 +30,7 @@ describe('Spacer', () => {
     expect(spacer.props.style.width).toBe(12)
   })
 
-  it('renders horizontally at overridden size vadsSpaceLg', () => {
+  it('renders horizontally at overridden size vadsSpaceLg (20)', () => {
     render(<Spacer size="lg" horizontal />)
 
     const spacer = screen.root
--- a/packages/components/src/components/Spacer/Spacer.test.tsx
+++ b/packages/components/src/components/Spacer/Spacer.test.tsx
@@ -12,7 +12,7 @@ describe('Spacer', () => {
     expect(spacer.props.style.width).toBe('auto')
   })
 
-  it('renders correctly at overridden size 20', () => {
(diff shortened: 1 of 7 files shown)


packages/components/src/components/Spacer/Spacer.tsx
```diff
@@ -1,25 +1,45 @@
...
+    | 'xs'
+    | 'sm'
+    | 'md'
+    | 'lg'
+    | 'xl'
+    | '2xl'
+    | '3xl'
+    | '4xl'
+    | '5xl'
+    | '6xl'
```
narin, line 20: @TimRoe Wanted to get your opinion on this. We want to lock the spacer size prop down to the available spacing tokens. I was thinking of either doing this by using the abbreviated spacer name as "variants" like above, OR locking it down to the explicit number values that match the spacing tokens.

### Abbreviated token  names
**Pros**
- User would not have to import spacing tokens or remember exact values
- If the spacing token values change, it wouldn't necessarily require changes to end user code or our Spacer code (they'd still have to account for adjustment in value)

**Cons**
- Token names don't mean much to the end user. They don't have much sense in what xs, s, med, etc mean
- Names don't match tokens exactly

### Number values
**Pros**
- User knows exact value, or can import spacing tokens

**Cons**
- Extra step in importing spacing tokens if doing so
- Less clear that its locked down to certain values
- If values change, will likely require code changes for end user

Curious if you had any thoughts or different approaches in mind.

packages/components/src/components/Spacer/Spacer.test.tsx
```diff
@@ -8,12 +8,12 @@ describe('Spacer', () => {
 
     const spacer = screen.root
 
-    expect(spacer.props.style.height).toBe(10)
+    expect(spacer.props.style.height).toBe(12)
     expect(spacer.props.style.width).toBe('auto')
   })
 
   it('renders correctly at overridden size 20', () => {
```
TimRoe, line 15:
```suggestion
  it('renders correctly at overridden size `lg` (20)', () => {
```

packages/components/src/components/Spacer/Spacer.test.tsx
```diff
@@ -8,12 +8,12 @@ describe('Spacer', () => {
```
TimRoe, line 8: Both `default size` test names (lines 6 and 24) need to be updated to:
`(12)`
or remove the size from the test

packages/components/src/components/Spacer/Spacer.test.tsx
```diff
@@ -27,11 +27,11 @@ describe('Spacer', () => {
     const spacer = screen.root
 
     expect(spacer.props.style.height).toBe('auto')
-    expect(spacer.props.style.width).toBe(10)
+    expect(spacer.props.style.width).toBe(12)
   })
 
   it('renders horizontally at overridden size 20', () => {
```
TimRoe, line 33:
```suggestion
  it('renders horizontally at overridden size `lg` (20)', () => {
```

packages/components/src/components/Spacer/Spacer.tsx
TimRoe: Of those two, prefer the `Abbreviated token names` route. Both cons seem minor to me. The design they're working from should be using the tokens so they don't need to care what the actual size is and lopping off `vadsSpacer` seems sensible.

I do wonder if we can use TS somehow to get the suggestion to show both. My initial thought is `enum`, but not sure if there's a way to do an enum where they wouldn't need to import it--which I think should be avoided. 

If  something with enum can't work without import, does it work to apply doc to the individual values? As in:
```
size?:
  /** 0 */
    | 'none'
  /** 4 */
    | '2xs'
```
or something like that.

Be cool if there was a way to communicate the value as well as the auto-fill abbreviated name, but if we can't find a way then as is seems optimal to me.

packages/components/src/components/Spacer/Spacer.tsx
```diff
@@ -1,25 +1,45 @@
 import { View } from 'react-native'
 import React, { FC } from 'react'
 
+import { spacing } from '@department-of-veterans-affairs/mobile-tokens'
+
 export type SpacerProps = {
-  /** Size of the spacer, default 10 */
-  size?: number
+  /** Size of the spacer, default sm */
```
TimRoe, line 7:
```suggestion
  /** Size of the spacer, default sm (12) */
```

packages/components/src/components/Spacer/Spacer.tsx
```diff
@@ -1,25 +1,45 @@
...
+ * @param size - Size of the spacer, default sm (12)
  * @param horizontal - True for horizontal spacing
  * @returns A non-visible component sized to provide appropriate spacing
  */
-export const Spacer: FC<SpacerProps> = ({ size = 10, horizontal = false }) => {
+export const Spacer: FC<SpacerProps> = ({
+  size = 'sm',
+  horizontal = false,
+}) => {
+  const index = `vadsSpace${size[0].toUpperCase()}${size.slice(1)}`
```
TimRoe, line 35: Potentially better named `key`?

packages/components/src/components/Spacer/Spacer.tsx
```diff
@@ -1,25 +1,45 @@
...
+  const spacerSize: number = spacing[index as keyof typeof spacing]
```
TimRoe, line 36: Side thought on this:

I thought this was kind of messy, but poking at it a bit I couldn't find a better way.

Was disappointed that:
```
  const index: keyof typeof spacing = `vadsSpace${size}`
  const spacerSize: number = spacing[index]
```
doesn't work because of the case on `size` (while agreeing it should remain lower case for consistency with our arguments/props generally).

Probably more effort than it's worth, but feels like there is _a way_ to simplify/improve this. But don't feel strongly enough to insist on it.

TimRoe (changes requested):
Marking request changes for the minor fixes to unit test names/doc.

Responded to your request for an opinion and [had another longer comment](https://github.com/department-of-veterans-affairs/va-mobile-library/pull/474#discussion_r1752898027), but neither requires changes if you don't see a fairly straightforward way to improve with the additional thoughts since the potential gains are marginal.

packages/components/src/components/Spacer/Spacer.tsx
narin: Separated out SpacerSize into a separate prop and added some JSDoc links which should come in handy.

Screenshot 2024-09-11 at 12 58 38 PM

packages/components/src/components/Spacer/Spacer.tsx
narin: The other option would to be to create a map which would be more type safe but less concise. Something like this:

```ts
const spacerSizeToSpacingKey = {
  'none': 'vadsSpaceNone',
  '2xs': 'vadsSpace2xs',
  'xs': 'vadsSpaceXs',
  'sm': 'vadsSpaceSm',
  'md': 'vadsSpaceMd',
  'lg': 'vadsSpaceLg',
  'xl': 'vadsSpaceXl',
  '2xl': 'vadsSpace2xl',
  '3xl': 'vadsSpace3xl',
  '4xl': 'vadsSpace4xl',
  '5xl': 'vadsSpace5xl',
  '6xl': 'vadsSpace6xl',
};
```

TimRoe (approved):

TimRoe:
Approved.

Last note is the PR doc doesn't show that AC:
`Storybook is updated with the ability to ONLY choose spacing tokens, remove custom input`
was completed, but trusting Storybook handled it.

PR #485, opened 2024-09-17
[Feature] Checkbox Implementation

narin:
## Description of Change

Implementation ticket for Checkbox component ([Figma design](https://www.figma.com/design/Zzt8z60hCtdEzXx2GFWghH/branch/mAMh8vyVgsevAOungfvGN6/%F0%9F%93%90-Component-Library---Design-System---VA-Mobile?node-id=1414-454&node-type=canvas&t=XqBHXGrMGvDIo65p-0))
- Required props
  - label
  - onPress
- Optional props
  - checked
  - description
  - descriptiona11y
  - hint
  - headerA11y
  - error
  - errorA11y
  - interderminate: shows indeterminate icon (overrides checked)
  - required: displays (*Required) text next to label
  - tile

### Notes:
- **For error text:** screen reader should announce "Error", then the provided error text
- Pressable area is the checkbox icon, label, description OR the entire tiled area when `tile={true}`
- Tested in the app by returning the new Checkbox component in VASelector when selector type is Checkbox or unspecified

#### Testing Packages

components
- 0.24.2-alpha.0
- 0.24.2-alpha.1

#### iOS

https://github.com/user-attachments/assets/282ce1fc-314c-4ddf-96e8-619167d2f565

#### iOS Voiceover

https://github.com/user-attachments/assets/04d320bd-691e-4819-bc25-fd5ee34e3972

#### Android

https://github.com/user-attachments/assets/315ead94-3567-4125-bda6-099ff2d2ebb0

#### Android Talkback

https://github.com/user-attachments/assets/e33f77f6-3207-42bd-8ef1-5b97b5750064

#### va-mobile-app

https://github.com/user-attachments/assets/d5ed9765-2959-4b16-bf34-86533e47fcc5

#### Web

https://github.com/user-attachments/assets/f780adfb-02f3-4f6f-99c5-a2c14e5df398

## Testing

- [x] Tested on iOS 
- [x] Tested on Android 
- [x] Tested on Web

## PR Checklist
Code reviewer validation:
- General
	- [x] Acceptance criteria: 
		- All satisfied _or_
- Code

## Publish

- [ ] Merge branch to `main`


packages/components/.storybook/web/preview.tsx
```diff
@@ -22,4 +22,7 @@ export const parameters = {
       return React.createElement(DocsContainer, currentProps)
     },
   },
+  viewport: {
+    defaultViewport: 'mobile2',
+  },
```
narin, line 27: Unrelated to ticket, but sets default viewport to mobile in Storybook Web. Let me know if you have any issues with this.

packages/components/src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,154 @@
...
+        {Story()}
+      </View>
+    ),
+  ],
+  parameters: {
+    design: [
+      {
+        name: 'Figma component overview',
+        type: 'figma',
+        url: 'REPLACE',
```
narin, line 23: Figma URLs/docs/etc to be refined in #456

packages/components/.storybook/web/preview.tsx
TimRoe: At face value, I didn't think I'd like it, but seems alright to me.

Only concern is if it might impact the iframes in the doc site. I don't think it will, but something to check once merged.

packages/components/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,264 @@
+import { Pressable, StyleProp, Text, View, ViewStyle } from 'react-native'
+import React, { FC } from 'react'
+
+import { spacing } from '@department-of-veterans-affairs/mobile-tokens'
+import { useTranslation } from 'react-i18next'
```
TimRoe, line 5:
```suggestion
import { spacing } from '@department-of-veterans-affairs/mobile-tokens'
import { useTranslation } from 'react-i18next'
import React, { FC } from 'react'
```

packages/components/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,264 @@
...
+
+  const baseTextStyle = {
+    fontFamily: 'SourceSansPro-Regular',
+    color: theme.vadsColorForegroundDefault,
+  }
+
+  /**
+   * Container styling
+   */
+
```
TimRoe, line 73: Very nitpicky, but might "block" things slightly better to remove the whitespace line between comment and first value within, while having the whitespace below to separate from the next block.

packages/components/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,263 @@
...
+  indeterminate,
+  onPress,
+  required,
+  tile,
+}) => {
+  const { t } = useTranslation()
+  const theme = useTheme()
+
+  const baseTextStyle = {
+    fontFamily: 'SourceSansPro-Regular',
```
TimRoe, line 65: Don't feel super strongly, but suggest reworking this slightly to have separate "fonts" so that the typography tokens can be dropped in more easily than having fontFamily split from fontSize/lineHeight and needing to shuffle around more later.

packages/components/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,263 @@
...
+      {required && <Text style={requiredStyle}>{` (*${t('required')})`}</Text>}
+    </Text>
+  )
+
+  /**
+   * Header
+   */
+
+  const headerProps = {
+    'aria-label': headerA11y || header,
```
TimRoe, line 172: Since it's applied directly to the text, not sure if we need the `||` or if it being undefined will automatically defer to the text; could potentially simplify here and other spots.

packages/components/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,263 @@
...
+
+  /**
+   * Hint
+   */
+
+  const hintProps = {
+    'aria-label': hintA11y || hint,
+    style: {
+      ...baseTextStyle,
+      color: theme.vadsColorForegroundSubtle,
```
TimRoe, line 191: Could also be simplified if doing the suggested rework above about centering on fonts so this isn't overwriting props from baseTextStyle and is just the font + color.

packages/components/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,263 @@
...
+  const _description = description && (
+    <>
+      <Spacer size="xs" />
+      <Text {...descriptionProps}>{description}</Text>
+    </>
+  )
+
+  return (
+    <ComponentWrapper>
+      <View style={containerStyle}>
```
TimRoe, line 243: Should this or the Pressable have `role="checkbox"`?

packages/components/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,263 @@
...
+        {_header}
+        {_hint}
+        {_error}
+        <Pressable
+          onPress={onPress}
+          style={tile ? tileStyle : pressableBaseStyle}
+          aria-checked={checked}
+          accessibilityState={{ checked }}
```
TimRoe, line 251: 1. Do we need both of these? Seems like they're doing the same thing
2. Do these need to be updated to also include logic for `intermediate` to be "mixed"? Currently if it was intermediate, the screen reader wouldn't indicate anything, right?

packages/components/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,263 @@
...
+  /**
+   * Icon
+   */
+
+  const iconViewStyle: ViewStyle = {
+    alignItems: 'center',
+    justifyContent: 'center',
+    paddingTop: 3,
+    paddingBottom: 3,
+  }
```
TimRoe, line 121: I didn't check, but think this will break with font scaling to maintain the checkbox centered with the text. Recommend following what we do in Alert/Link/Snackbar (including updates in `main`, but not yet your branch to depend on the font).

packages/components/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,263 @@
...
+
+  const _icon = (
+    <View style={iconViewStyle}>
+      <Icon
+        name={
```
TimRoe, line 126: Might require changing `_icon` to a function to continue updating as expected, but suggest pulling out the icon name (and possibly fill) to not have a more complicated than single line ternary within the display logic.

packages/components/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,263 @@
...
+  }
+
+  const _description = description && (
+    <>
+      <Spacer size="xs" />
+      <Text {...descriptionProps}>{description}</Text>
+    </>
+  )
+
+  return (
```
TimRoe, line 241: Minor and nitpicky, but noting that the the order above not matching the order here is a little disorienting (probably more from looking at it in browser without Cmd + clicking around).

packages/components/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,263 @@
...
+    <ComponentWrapper>
+      <View style={containerStyle}>
+        {_header}
+        {_hint}
```
TimRoe, line 245: Not sure I want to see it changed, but just throwing out that it's more difficult to follow the Spacer use having it nested within the various pieces. But it also may be obnoxious to have all these separated by `_header && _hint ? <Spacer > : null`.

I definitely like the general principle of splitting out these named parts though as it helps to see the high level pertinent structure.

packages/components/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,263 @@
...
+
+import { ComponentWrapper } from '../../wrapper'
+import { Icon } from '../Icon/Icon'
+import { Spacer } from '../Spacer/Spacer'
+import { useTheme } from '../../utils'
+
+export type CheckboxProps = {
+  /** Primary text for checkbox */
+  label: string
+  /** OnPress function. Pass a function that alters `checked` state */
```
TimRoe, line 14: Reads as maybe too prescriptive documentation? Could also alter indeterminate and other things on the app level.

```suggestion
  /** OnPress logic to alter `checked` state or other behavior associated with the checkbox */
```
?

packages/components/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,263 @@
...
+  onPress: () => void
+  /** Optional boolean that determines whether checkbox appears checked */
+  checked?: boolean
+  /** Optional description that appears below label */
+  description?: string
+  /** Optional accessibility override for description text  */
+  descriptionA11y?: string
```
TimRoe, line 21: With how much a11y label overriding we have going here, I wonder if a better solution might be something like:
```
type textWithA11y = {
  text: string
  a11y: string
}
```

and typing this like:
```suggestion
  description?: string | textWithA11y
```

Would add some complexity using the values though, might be worth playing with locally and seeing if it shakes out to be beneficial to simplify the props from being so cluttered with text + a11y combos.

packages/components/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,263 @@
...
+  /** Optional hint text. Appears below header */
+  hint?: string
+  /** Optional accessibility override for hint text  */
+  hintA11y?: string
+  /** Optional error text. Non-null value styles checkbox in error state */
```
TimRoe, line 26:
```suggestion
  /** Optional error text. Applies error styling to checkbox if present */
```

packages/components/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,263 @@
...
+  errorA11y?: string
+  /** Optional header text */
+  header?: string
+  /** Optional accessibility override for header  */
+  headerA11y?: string
+  /** Optional indeterminate boolean. Overrides checked styling */
+  indeterminate?: boolean
+  /** Optional accessibility override for label  */
+  labelA11y?: string
+  /** Optional required boolean that denotes a label as (*Required) */
```
TimRoe, line 38: "Optional required" feels weird to read even if it makes sense with context.

```suggestion
  /** True to append (*Required) suffix to label */
```

packages/components/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,263 @@
...
+  required?: boolean
+  /** Optional tile styling */
+  tile?: boolean
+}
```
TimRoe, line 42: Feel like we probably lack consistency around it and it's just a preference thing, but just noting that the "Optional" nature of these props is covered by the typing and they could be more direct in just saying what the string does or "True to ..." for booleans.

packages/components/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,263 @@
...
+          indeterminate
+            ? 'IndeterminateCheckBox'
+            : checked
+              ? 'CheckBox'
+              : 'CheckBoxOutlineBlank'
+        }
+        fill={
+          checked
+            ? theme.vadsColorFormsForegroundActive
+            : theme.vadsColorFormsBorderDefault
```
TimRoe, line 136: I glanced at Figma, but is it a bug that the `indeterminate` status is styling to be gray instead of blue? Seems like it, but can't point to anything definite.

```suggestion
          checked || indeterminate
            ? theme.vadsColorFormsForegroundActive
            : theme.vadsColorFormsBorderDefault
```
?

TimRoe (changes requested):
Marking request changes to denote end of first pass (besides the stories, which should be fine and have a separate ticket).

Looks mostly good, quite a few comments/suggestions for improvement. I think the notable needed fixes are:
- Indeterminate icon color bugged
- Icon alignment with text is likely bugged (didn't verify) with font scaling, should probably mimic Alert/Link/Snackbar solution
   - Wonder if this is reaching prevalence that we should have a function in the Icon file that takes the appropriate text lineHeight it needs to align with and is a function that returns the style
      - Or maybe even a component? IconAlignedWithText or something that returns the whole View and takes props for IconProps, text size it's aligned with, and any other custom View props?
      - Can probably split to a separate ticket if it makes sense to you

packages/components/src/components/Checkbox/Checkbox.stories.tsx
```diff
@@ -0,0 +1,154 @@
...
+    hint,
+    hintA11y,
+    indeterminate,
+    label,
+    labelA11y,
+    required,
+    tile,
+  } = props
+
+  const [checked, setChecked] = useState(false)
```
TimRoe, line 59: Wonder if doing the same with indeterminate will fix the weirdness observed in the one recording you had turning it on.

narin:
- Updated prop comments and removed redundant use of “Optional”
- Separated out typography styles instead of using baseTextStyle to prepare for typography tokens
- Adjusted comment whitespace and sorted sections by display order
- Fixed icon styling to maintain centered style for font scaling
- Extracted icon prop logic from display logic
- Simplified `aria-label`
- Removed `aria-checked` since `accessibilityState` serves the same purpose.
  - Added logic for indeterminate to set as `’mixed'`

packages/components/src/components/Checkbox/Checkbox.tsx
TimRoe: > I think this component probably has some of the most individual text props we'll encounter and it may be okay as is.

Seems like Radio will probably be the same and other form components will have at least 2 (label, error) if not more.

Agree with the general notion it should be applied to existing components, but could also see a caveat of `only applies to components with more than one "text" value` which would mean just this and Alert (and technically Segmented control, but that's using arrays for the labels and corresponding a11y so it's still "1" in a sense).

I wonder if we should start making a list of items to target/consider as part of going version 1.0.0 where we can feel free to make widespread breaking changes when it's expected as opposed to the current situation where flagship is largely hopping component-to-component version with no regard to what else we're up to.

All that said, this would be mostly for our benefit and not do much for consumers so maybe it's fine, if a bit obnoxious to look at.

packages/components/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,274 @@
...
+    alignItems: 'flex-start',
+  }
+
+  const tileStyle: ViewStyle = {
+    ...pressableBaseStyle,
+    borderWidth: 2,
+    borderRadius: spacing.vadsSpace2xs,
+    paddingTop: spacing.vadsSpaceSm,
+    paddingBottom: spacing.vadsSpaceSm,
+    paddingLeft: spacing.vadsSpaceSm,
```
TimRoe, line 103: Could be
```suggestion
    padding: spacing.vadsSpaceSm,
```
as specific sides should override the general rule.

packages/components/src/components/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,274 @@
...
+      {required && <Text style={requiredStyle}>{` (*${t('required')})`}</Text>}
+    </Text>
+  )
+
+  /**
+   * Icon
+   */
+  const iconViewStyle: ViewStyle = {
+    paddingTop: 3,
+    paddingBottom: 3,
```
TimRoe, line 206: Should no longer be needed as the other props center it within the minHeight that matches the font lineHeight
```suggestion

```

TimRoe:
> label has to be before icon to reference label lineHeight

Don't feel strongly that it's updated, but my suggestion to match other components was such that at the top there'd be something like:
```
const font = {fontFamily: ..., fontSize: ..., lineHeight: ...}
const fontBold = ...
const fontHint = ...
```
where when we had the typography tokens that entire block could be deleted and updating the `...font`, `...fontBold`, `...fontHint` to the tokens--in that way, the needed lineHeight would be defined as it will when we have the tokens.

TimRoe:
> Removed `aria-checked` since `accessibilityState` serves the same purpose.

Doesn't matter currently, but just making a note that `aria-checked` is the newer/more web-like implementation. Not sure there's any plans to deprecate anytime soon at the RN level, but know Tim W had pushed us towards using the newer ones because I think Web Storybook issues console warnings for using the older ones.

TimRoe (approved):
Some additional suggestions/comment, but nothing critical. Approved.

Issue #448, opened 2024-08-19
DS - Add Spacing Tokens to Spacer Component


kellylein:
## Description 
As part of this story we want to:   

- Update the existing Spacer component to utilize the newly added spacing tokens from #447

## Design Notes
- [Notes](https://www.figma.com/design/XQwpzcNSGyGJN4nXDPoEj0/Design-Tokens---%F0%9F%94%8D-Design-System---VA-Mobile?node-id=225-1140&t=vQm9WlaU68BEyqVD-4)
- [JSON export from Figma](https://www.figma.com/design/XQwpzcNSGyGJN4nXDPoEj0/Design-Tokens---%F0%9F%94%8D-Design-System---VA-Mobile?node-id=224-11312&t=vQm9WlaU68BEyqVD-4)
- [Spacing tokens](https://www.figma.com/design/XQwpzcNSGyGJN4nXDPoEj0/Design-Tokens---%F0%9F%94%8D-Design-System---VA-Mobile?node-id=224-11338&t=vQm9WlaU68BEyqVD-4)

<img width="657" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBd2F6Qmc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--b569c8ba8426e7303fa2e7205a506f735b8848b7/image.png" alt="image.png" />

## Acceptance Criteria
  

-  [ ] Spacer component utilizes spacing tokens
-  [ ] Storybook is updated with the ability to ONLY choose spacing tokens, remove custom input

## Notes & Open Questions

- Dependencies/Roadblocks:
    - Any internal/external dependencies?
    - Test accounts needed?
- Does this require QA?
    
- Dev Notes: 

## Ticket Checklist

- [x] Acceptance criteria defined
- [x] Labels added (front-end, back-end, feature)
- [x] Linked to an Epic

PR #487, opened 2024-09-17
[Code Upkeep] Update components to use spacing tokens

TimRoe:
## Description of Change

- Updated Alert, Button, Link, Segmented control, and Snackbar to use the spacing tokens
   - ~~Loading indicator did not require changes as it used the default spacing size which was updated by #448~~
   - Loading indicator updated during PR to use `xs` (8) spacers between icon/text and text/children content
- Fixed a minor design inconsistency with Segmented control where it had no horizontal padding so segment labels could get very close before wrapping
   - Created #491 to address further improvements
- Fixed a minor bug noted with Snackbar where it was applying the `gap` property to both row and column flex so it had minor 5 extra padding between message and buttons causing it to wrap slightly sooner than it should've
- Reworked setting of height for icon containers to align with the text (Alert, Link, Snackbar) to rely directly on the font's `lineHeight` property instead of a "magic number"

#### Testing Packages

- [0.24.2-alpha.2](https://www.npmjs.com/package/@department-of-veterans-affairs/mobile-component-library/v/0.24.2-alpha.2)

## Testing

While working iteratively through the components, checked each platform was incorporating spacing changes (or same but using tokens) as expected.

Smoke tested in app everything seemed good in iOS. Only checked components actually used in the app currently (not Loading indicator, not Snackbar). Everything tested fine except that the app is using Spacer [directly in one spot](https://github.com/department-of-veterans-affairs/va-mobile-app/blob/97d6a9b5021e0ce1b440b7cef973d99fc2f4542e/VAMobile/src/screens/HealthScreen/Appointments/AppointmentTypeComponents/SharedComponents/AppointmentLocation.tsx#L208) which crashed because now we do not allow custom sizes and they were sending "20" which then chokes on .toUpperCase because it's a number, not a string. Updating it to have size `lg` behaved as expected.

- [x] Tested on iOS 
- [x] Tested on Android 
- [x] Tested on Web

## PR Checklist
Code reviewer validation:
- General
	- [ ] PR is linked to ticket(s)
	- [ ] PR has `changelog` label applied if it's to be included in the changelog
	- [ ] Acceptance criteria: 
		- All satisfied _or_
	- [ ] Above PR sections adequately filled out
	- [ ] If any breaking changes, in accordance with the [pre-1.0.0 versioning guidelines](https://github.com/department-of-veterans-affairs/va-mobile-library#versioning-policy): a CU ticket has been created for the VA Mobile App detailing necessary adjustments with the package version that will be published by this ticket
- Code
	- [ ] Tests are included if appropriate (or split to separate ticket)
	- [ ] New functions have proper TSDoc annotations

## Publish

- [x] Merge branch to `main`
- [x] Verify that [Check Component Integrations](https://github.com/department-of-veterans-affairs/va-mobile-library/actions/workflows/check-component-integrations.yml) workflow ran successfully
- [x] [Publish new version](https://github.com/department-of-veterans-affairs/va-mobile-library/actions/workflows/publish.yml) after merging

### DIFF
--- a/packages/components/src/components/Alert/Alert.tsx
+++ b/packages/components/src/components/Alert/Alert.tsx
@@ -7,6 +7,7 @@ import {
   ViewStyle,
   useWindowDimensions,
 } from 'react-native'
+import { spacing } from '@department-of-veterans-affairs/mobile-tokens'
 import React, { FC, useState } from 'react'
 
 import { BaseColor, useColorScheme, useTheme } from '../../utils'
@@ -88,18 +89,24 @@ export const Alert: FC<AlertProps> = ({
     setExpanded(!expanded)
   }
 
-  // TODO: Replace with sizing/dimension tokens
-  const Sizing = {
-    _8: 8,
-    _10: 10,
-    _12: 12,
-    _16: 16,
-    _20: 20,
-    _24: 24,
-    _30: 30,
-  }
   const contentColor = AlertContentColor()
   let backgroundColor, borderColor, iconName: IconProps['name']
+
+  // TODO: Replace with typography tokens
+  const headerFont: TextStyle = {
+    color: contentColor,
+    fontFamily: 'SourceSansPro-Bold',
+    fontSize: 20,
+    lineHeight: 30,
+  }
+
+  // TODO: Replace with typography tokens
+  const descriptionFont: TextStyle = {
+    color: contentColor,
+    fontFamily: 'SourceSansPro-Regular',
+    fontSize: 20,
+    lineHeight: 30,
+  }
 
   switch (variant) {
     case 'info':
@@ -127,9 +134,9 @@ export const Alert: FC<AlertProps> = ({
   const contentBox: ViewStyle = {
     backgroundColor: backgroundColor,
     borderLeftColor: borderColor,
-    borderLeftWidth: Sizing._8,
-    padding: Sizing._20,
-    paddingLeft: Sizing._12, // Adds with borderLeftWidth for 20
+    borderLeftWidth: spacing.vadsSpaceXs,
+    padding: spacing.vadsSpaceLg,
+    paddingLeft: spacing.vadsSpaceSm, // Adds with borderLeftWidth to be `Lg` (20)
     width: '100%', // Ensure Alert fills horizontal space, regardless of flexing content
   }
 
@@ -137,7 +144,7 @@ export const Alert: FC<AlertProps> = ({
     flexDirection: 'row',
     // Below keeps icon aligned with first row of text, centered, and scalable
     alignSelf: 'flex-start',
-    minHeight: Sizing._30 * fontScale,
+    minHeight: headerFont.lineHeight! * fontScale,
     alignItems: 'center',
     justifyContent: 'center',
   }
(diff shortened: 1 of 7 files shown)


packages/components/src/components/Alert/Alert.tsx
```diff
@@ -127,25 +134,25 @@ export const Alert: FC<AlertProps> = ({
   const contentBox: ViewStyle = {
     backgroundColor: backgroundColor,
     borderLeftColor: borderColor,
-    borderLeftWidth: Sizing._8,
-    padding: Sizing._20,
-    paddingLeft: Sizing._12, // Adds with borderLeftWidth for 20
+    borderLeftWidth: spacing.vadsSpaceXs,
```
jessicawoodin, line 137: Spacing tokens should not be used for border widths. In the future, we may create tokens specifically for borderWidth.

jessicawoodin:
Overall, this looks great! I didn't notice any issues with the Button, Link, Snackbar, or Segmented Control. I left one comment for the Alert. I didn't see any changes to the Loading Indicator, but when I was reviewing that, I realized my decision to change the padding from 10 to 12 (instead of 8) was not consistent with VADS. Sorry about that! Would we be able to change the spacing between the icon and loading message to xs(8) instead of sm(12)? Thank you!

packages/components/src/components/Alert/Alert.tsx
TimRoe: This is a special case because how react native handles things. If you note the other properties here, the overall padding is 20, but the left padding is 12 because it's additive with the border width to achieve 20 padding on the left side so for practical purposes this is effectively padding. It's just special padding to have the different color sidebar on the Alert.

I think it makes sense until there's a dedicated token. Even then, this is being used as padding more than a border since it's only on one side. 

Unless Narin thought otherwise, inclined to leave it as spacing.

TimRoe:
> I didn't see any changes to the Loading Indicator, but when I was reviewing that, I realized my decision to change the padding from 10 to 12 (instead of 8) was not consistent with VADS. Sorry about that! Would we be able to change the spacing between the icon and loading message to xs(8) instead of sm(12)?

Can do, but should the spacing between the loading text and any custom content remain 12 too or also shrink to 8?

jessicawoodin:
> > I didn't see any changes to the Loading Indicator, but when I was reviewing that, I realized my decision to change the padding from 10 to 12 (instead of 8) was not consistent with VADS. Sorry about that! Would we be able to change the spacing between the icon and loading message to xs(8) instead of sm(12)?
> 
> Can do, but should the spacing between the loading text and any custom content remain 12 too or also shrink to 8?

That should also change to 8.

packages/components/src/components/Alert/Alert.tsx
narin: I ran into the same thing with the red left border for the error state in Checkbox. As Tim mentioned, in figma, borders are inset and its width is included in the padding calculation whereas in RN, they are separate.

I don't feel super strongly but I think I'm also leaning towards leaving this until we have dedicated tokens as it does serve as padding in a way. Also best to use tokens instead of absolute numbers when possible.

packages/components/src/components/SegmentedControl/SegmentedControl.tsx
```diff
@@ -35,9 +36,11 @@ type SegmentProps = {
 
 const Segment = styled(Pressable)<SegmentProps>`
   border-radius: 8px;
-  padding-vertical: 7px;
+  padding-horizontal: ${spacing.vadsSpace2xs}px;
+  padding-vertical: ${spacing.vadsSpaceXs}px;
```
narin, line 40: Shouldn't the padding be 2xs (4) for both horizontal and vertical? 

Screenshot 2024-09-20 at 10 07 16 AM

narin (changes requested):
Noticed 1 difference between figma and code for the SegmentedControl padding but otherwise looks great.

packages/components/src/components/SegmentedControl/SegmentedControl.tsx
TimRoe: Figma didn't get updated (cc: @jessicawoodin ), but we changed it in the [thread for the ticket](https://dsva.slack.com/archives/C07KFQ3U6QH/p1726684056537069?thread_ts=1726599619.164929&cid=C07KFQ3U6QH):
> But to answer your question, sure, let's go with xs/8 for the top/bottom padding. And I'll set the left/right padding to 2xs/4.

due to a misalignment of the existing component as implemented (5 vertical padding in Figma, 7 vertical padding in the implemented component) so adjusted to do what's the closest to current--and also kept things simpler by ensuring the touch target didn't become smaller than the 44 pixels tall.

TimRoe: Also, could re-evaluate on #491 if desired. There were some other alignment issues with Segmented control between implemented component and Figma that could likely be squared away a bit. As our first component, it was a bit of a proving ground for other things and leaned on the existing app implementation perhaps more than it should've.

narin (approved):

PR #503, opened 2024-09-25
[Feature] CheckboxGroup

narin:
## Description of Change

This PR contains the CheckboxGroup implementation along with some major refactoring to make common props and text elements more reusable. Open to suggestions for naming and folder structure
### Text Types
  -  Created 3 new reusable types in `types/common.ts` that should replace the need for separate "*A11y" props for components: `TextA11y` and `AccessibleText`
    - `TextWithA11y` is an object of type `{ text: string, a11y?: string }`
    - `StringOrTextWithA11y` can be of type `string | TextA11y` which allows for simple strings or text with a11y overrides
    - `TextWithA11yAndValue` is `TextWithA11y` but with a `value` field added. Used for CheckboxGroup and likely RadioButton
- Added some helper utils, `getDisplayText` and `getA11yText` to make getting text and a11y values from `StringOrTextWithA11y` easier

### Form Types
 -  Created `types/form.ts` to contain props that will be common across many form elements. Went back and forth on whether types should live within their related component folders or in this `types/` folder

### Form Components
  - Created a new folder `src/components/Forms` to house individual and common form components
  - Extracted Header, Hint, Error, Label, and Description into components that live in a new file called `Forms/FormText.tsx`
  - Updated Checkbox to use common text elements
  - CheckboxGroup
      - Similar to `StringOrTextWithA11y`, wanted to make the `items` passed to CheckboxGroup flexible, so it accepts an an array of strings or an array of `TextWithA11yAndValue`
      - If no values are provided, uses `item` as the value or `item.text` if `item.value` isn't provided
  - **Misc**
    - Noticed that the viewport change from Checkbox was resulting in some weird behavior on web where I couldn't access the addons. Reverted viewport to default
    - Added a setting that will sort controls alphabetically and put required props first
    - Fixed a bug where screen reader wasn't reading "Required" if label or header has "(*Required)" suffix

### Screenshots/Video

<details><summary>iOS</summary>

https://github.com/user-attachments/assets/ff498f5d-9724-4a39-9cfa-bd18dec198b1
</details>

<details><summary>iOS Voiceover</summary>

https://github.com/user-attachments/assets/3dc1e194-291d-45f7-bf1d-df1023915e74
</details>

<details><summary>Android</summary>

https://github.com/user-attachments/assets/c62e5c50-84b1-4531-b17a-b3c2dc75f6cb
</details>

<details><summary>Android Talkback</summary>

https://github.com/user-attachments/assets/d18de16b-98dc-4f79-814b-c6a202668fd1
</details>

<details><summary>Web</summary>

https://github.com/user-attachments/assets/a1390711-bd4c-4f46-847f-237d5f68cf5f
</details>

<details><summary>va-mobile-app</summary>

https://github.com/user-attachments/assets/2c4dc981-376d-4045-9bcb-19215bbed0a4
</details>

## Testing

- [x] Tested on iOS 
- [x] Tested on Android 
- [x] Tested on Web

## PR Checklist
Code reviewer validation:
- General
	- [x] Acceptance criteria: 
		- All satisfied _or_
- Code

## Publish

If changes warrant a new version [per the versioning guidelines](https://github.com/department-of-veterans-affairs/va-mobile-library/blob/main/documentation/versioning.md) and the PR is approved and ready to merge:
- [ ] Merge branch to `main`


packages/components/src/components/Forms/Checkbox/Checkbox.tsx
```diff
@@ -0,0 +1,154 @@
...
+
+  return (
+    <ComponentWrapper>
+      <View style={containerStyle}>
+        {header && (
+          <>
+            <Header text={header} />
+            <Spacer size="xs" />
+          </>
+        )}
```
narin, line 118: Don't love this conditional rendering. Was going back and forth on whether the text components should accept an undefined value for `text` and return `null` in those cases which would clean this up a bit.

packages/components/src/components/Forms/FormText.tsx
```diff
@@ -0,0 +1,167 @@
...
+  lineHeight: 30,
+}
+
+const fontDescription = {
+  fontFamily: fontRegular,
+  fontSize: 16,
+  lineHeight: 22,
+}
+
+/**
```
narin, line 48: Could potentially split these all into separate files instead

packages/components/src/components/Checkbox/Checkbox.tsx
narin, line 1: Moved to `/Forms` folder. Not sure why this file is showing as deleted whereas `Checkbox.stories.tsx` is showing as renamed.

TimRoe: No real idea, but would guess some arbitrary level of diff difference by GitHub triggering it to not consider it "same-y" enough through the move.

packages/components/src/components/Forms/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+        {error && (
+          <>
+            <Error text={error} />
+            <Spacer size="xs" />
+          </>
+        )}
+        {items.map((item, index) => {
+          const value =
+            typeof item === 'object' ? item.value || item.text : item
```
TimRoe, line 85: I'm finding this very confusing even after looking for a good amount of time. I assume it may have something to do with the form wrapper, but this "value" prop seems incredibly ambiguous. Is it supposed to be a key/id that can be used to identify which Checkbox item is being acted on instead of just the label?

Maybe related to the general not understanding, but it further seems unclear why this is needed vs the selectedItems depending solely on the labels which should always be present.

packages/components/src/types/common.ts
```diff
@@ -0,0 +1,13 @@
+export type TextWithA11y = {
+  /** Text to display */
+  text: string
+  /** Accessibility override for text */
+  a11y?: string
```
TimRoe, line 5: This should likely be named `a11yLabel` or `label` or something to facilitate inclusion of other a11y overrides that can apply to text (hint, value, etc.) later.

packages/components/src/types/common.ts
```diff
@@ -0,0 +1,13 @@
...
+}
+
+export type TextWithA11yAndValue = TextWithA11y & {
+  /** Value that may differ from provided text */
+  value?: string | number
+}
+
+export type StringOrTextWithA11y = string | TextWithA11y
```
TimRoe, line 13: I'm not sure this adds value as a named type. 

On the one hand, it provides a "do not repeat yourself" value. On the other hand, it's so specific that it couldn't be flexible in the future without renaming anyway so I'm not sure it adds meaningful value vs just having `string | TextWithA11y` typed on the props themselves.

TimRoe: Caveat: it may make more sense if we were exporting it for consumer consumption so when they were nesting things and passing along the typing they could use one thing instead of mimicking our `string | TextWithA11y`.

packages/components/src/types/common.ts
```diff
@@ -0,0 +1,13 @@
...
+  /** Text to display */
+  text: string
+  /** Accessibility override for text */
+  a11y?: string
+}
+
+export type TextWithA11yAndValue = TextWithA11y & {
+  /** Value that may differ from provided text */
+  value?: string | number
+}
```
TimRoe, line 11: Perhaps tied to my other comment on not really understanding this property, but at face value it seems like this should live with the CheckboxGroup and perhaps move to `FormText.tsx` for the Radio button component when it's actually shared. Outside those use cases, it's unclear to me what purpose this would have as a shared typing across most components.

packages/components/src/utils/accessibility.ts
```diff
@@ -1,2 +1,16 @@
+import { StringOrTextWithA11y } from '../types/common'
+
 // Export related hooks
 export { useIsScreenReaderEnabled } from './hooks/useIsScreenReaderEnabled'
+
+/**
+ * Returns text that should be displayed on the screen
+ */
+export const getDisplayText = (text: StringOrTextWithA11y): string =>
```
TimRoe, line 9: Not a hook and exported:
```suggestion
export const GetDisplayText = (text: StringOrTextWithA11y): string =>
```

packages/components/src/utils/accessibility.ts
```diff
@@ -1,2 +1,16 @@
...
+  typeof text === 'string' ? text : text.text
+
+/**
+ * Returns a11y override text or just text if a11y isn't provided
+ */
+export const getA11yText = (text: StringOrTextWithA11y): string =>
```
TimRoe, line 15: Not a hook and exported:
```suggestion
export const GetA11yText = (text: StringOrTextWithA11y): string =>
```

packages/components/src/components/Forms/FormText.tsx
```diff
@@ -0,0 +1,167 @@
+import { Text } from 'react-native'
```
TimRoe, line 1: After mulling it over a bit, I don't think I'm a fan of the Forms folder vs keeping the components file structure nice and flat of one folder, one component, no nesting/depth. Especially since some of the "forms" components are not necessarily intrinsically tied to form use cases. For example: secure messaging using text inputs and text boxes or claims using an individual Checkbox for acknowledgement or Rx Refill sort/filter where none feels like a "form" per se.

Not sure I have any good suggestions for renaming this file to be more generalized (maybe something like `sharedElements`?), but am thinking it preferable to flatten Checkbox/CheckboxGroup back into the `components` folder and shuffle this off to a new utils file.

packages/components/src/components/Forms/FormText.tsx
TimRoe: One file seems fine to me since most of what's in here besides this will be deleted with Typography tokens and it isn't a huge list that feels unwieldy.

packages/components/src/components/Forms/FormText.tsx
```diff
@@ -0,0 +1,167 @@
...
+import React, { FC } from 'react'
+
+import { StringOrTextWithA11y } from '../../types/common'
+import { getA11yText, getDisplayText } from '../../utils'
+import { useTheme } from '../../utils'
+import { useTranslation } from 'react-i18next'
```
TimRoe, line 7: Library/local import mixing.

Wonder if we can update our eslint or TS config to handle this for us--I feel like whatever configuration flagship has does automatically handle split sorting libraries then local.

packages/components/src/types/forms.ts
```diff
@@ -0,0 +1,29 @@
...
+  /** Optional error text. If present, applies error styling to element */
+  error?: StringOrTextWithA11y
+  /** Header text */
+  header?: StringOrTextWithA11y
+  /** Hint text. Appears below header */
+  hint?: StringOrTextWithA11y
+  /** True to append (*Required) suffix to element */
+  required?: boolean
+  /** True to apply tile styling */
+  tile?: boolean
```
TimRoe, line 16: Will this be common to any besides the Checkbox(Group) and Radio? Or should it be in the below typing?

packages/components/src/components/Forms/FormText.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+export type FormTextProps = {
```
TimRoe, line 9: Seems pretty purpose made, potentially:
```suggestion
type formTextProps = {
```

packages/components/src/components/Forms/Checkbox/Checkbox.tsx
TimRoe: I'm in favor of moving it into the components and having basically the first line be `if (!text) return null`. 

Seems like a reasonable assumption that they are conditional on the presence of the value, albeit maybe a little less intuitive to track down for fresh eyes.

I am on the fence about then whether you move the Spacer with that or do a line of `{header && hint ? <Spacer /> : null}` between each.

packages/components/src/components/Forms/CheckboxGroup/CheckboxGroup.tsx
narin: The FormWrapper does use values for Radio Buttons, but I'm thinking beyond that to common use cases in forms. The app doesn't currently use any CheckboxGroups, but the options that show up in them or in Radio Buttons are likely to be a mapping of some list of values or IDs from an API, with separate display text.  Ultimately, these values are what are going to be passed back to the API in a form submit, not the labels. 

While it'd be simpler on our end to use labels passed to keep track of `selectedItems`, this would require an extra step on the user's end to then get the corresponding value to the selected label we pass them before submitting. So while it's a little more complex for us to support both, I think it'd be more convenient for devs.

packages/components/src/utils/accessibility.ts
TimRoe: Also probably rename to GetA11yLabel corresponding to other comment on TextWithA11y type.

packages/components/src/types/common.ts
TimRoe: Seeing how many places it's used working through the rest of the review--perhaps worth keeping even if not exported.

packages/components/src/components/Forms/FormText.tsx
```diff
@@ -0,0 +1,167 @@
...
+export const Description: FC<FormTextProps> = ({ text }) => {
+  const theme = useTheme()
+
+  const textStyle = {
+    ...fontDescription,
+    color: theme.vadsColorForegroundDefault,
+  }
+
+  return (
+    <Text aria-label={`, ${getA11yText(text)}`} style={textStyle}>
```
TimRoe, line 163:
```suggestion
    <Text aria-label={`${getA11yText(text)}`} style={textStyle}>
```
?

narin: Without the comma, screen readers won't pause between reading the Label in Description since the Description is nested with Label

packages/components/src/components/Forms/CheckboxGroup/CheckboxGroup.stories.tsx
```diff
@@ -0,0 +1,113 @@
...
+  component: CheckboxGroup,
+  decorators: [
+    (Story) => (
+      <View style={{ flex: 1, justifyContent: 'center', alignItems: 'center' }}>
+        {Story()}
+      </View>
+    ),
+  ],
+  parameters: {
+    design: [
```
TimRoe, line 19: Just adding a note that we've applied these very inconsistently on existing components and looking like one old case, Segmented control, has broken over time as Figma content got shuffled around.

@jessicawoodin, did you care if the Storybook pointed to Figma? I wonder if we're getting a bit duplicative here as Storybook does point to the doc site which includes both Storybook and Figma. General worries about ongoing maintenance as we know _someday_ Figma should move to VA's instance and this would be another spot to update.

For reference, but [Button is working to demo](https://department-of-veterans-affairs.github.io/va-mobile-library/?path=/story/button--primary) how it appears as the "Design" tab in the bar with the controls/actions:
image

packages/components/src/components/Forms/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+export type CheckboxGroupProps = FormElementProps & {
+  /** Array of checkbox options. Can be an array containing strings or objects if values or a11y overrides are needed */
+  items: string[] | TextWithA11yAndValue[]
+  /** Callback function that receives an updated array of selected values when checkboxes are pressed */
+  onSelectionChange: (selected: (string | number)[]) => void
+  /** Array of the labels or values (if provided) of currently selected checkboxes */
+  selectedItems: (string | number)[]
+}
+
```
TimRoe, line 21: Should we have additional documentation here explaining how to configure `onSelectionChange` and `selectedItems` with useState? When used outside FormWrapper at least since I assume that's baked in for it.

narin:
### RE: Naming Conventions
Since many of the comments are related, adding a general comment here. I think we've been following certain patterns pretty consistently so far but we should probably build upon our prop naming convention docs and come up with some "official" naming conventions for our repo. 

Most TS resources I've come across recommend UpperCamelCase for TS types and camelCase for functions, regardless of whether they are exported or not. While there's no official TS style guide, here are resources from [Google](https://google.github.io/styleguide/tsguide.html#naming-rules-by-identifier-type) and the [typescript repo](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#names). Of course we can come up with our own conventions and enforce them as strictly or as loosely as we'd like, but as  a default I'm leaning towards sticking to what's common in the community.

packages/components/src/types/common.ts
```diff
@@ -0,0 +1,13 @@
+export type TextWithA11y = {
```
TimRoe, line 1: Thoughts on an `index.ts` for the types folder? Making them all available as a group instead of per file basis.

packages/components/src/components/Forms/CheckboxGroup/CheckboxGroup.tsx
TimRoe: I think at a minimum we need something explaining this property better. "Value that may differ from provided text" is super vague and, with your clarification, I'm still not sure I follow the use case for this beyond CheckboxGroup and Radio.

Even that may be a stretch because I feel the uses are different:
- Radio:
    - Select one for one component
    - You pick one of many options (values) to store as data for one key/id/index/whatever in the app
- CheckboxGroup:
    - Select many for one or many component
        - You pick multiple of many options (values) to store as data for one key/id/index/whatever in the app OR 
        - You pick multiple of many options (values) to store as data for multiple separate key/id/index/whatever's in the app (each checkbox corresponds to app data)
    - On top of that, these are simultaneously keys themselves because they in turn store data: true/false of whether it is checked
    - So here these are simultaneously keys and values: keys for the item in the list (as being used from the component perspective), values for some app-level key or keys that cares about which boxes are checked

So I guess in fully thinking through it, they are more values than not coming from the consumer, but it still seems muddy in my mind that it's logical for CheckboxGroup/Radio to share, let alone potentially other components as well.

Maybe the answer is "trust me" from your research around FormWrapper, but as of this ticket I have skepticism we aren't trying to oversimplify saying this makes sense to be global vs just forms vs local to CheckboxGroup.

TimRoe:
Probably just relative to when recorded, but noting that the Google Talkback video doesn't align with the stories as coded in terms of a11y. I'm assuming that's just due to config at recording time and not a bug since I don't see anything amiss in the code.

TimRoe:
> ### RE: Naming Conventions
> Since many of the comments are related, adding a general comment here. I think we've been following certain patterns pretty consistently so far but we should probably build upon our prop naming convention docs and come up with some "official" naming conventions for our repo. 
>
> Most TS resources I've come across recommend UpperCamelCase for TS types and camelCase for functions, regardless of whether they are exported or not. While there's no official TS style guide, here are resources from [Google](https://google.github.io/styleguide/tsguide.html#naming-rules-by-identifier-type) and the [typescript repo](https://github.com/Microsoft/TypeScript/wiki/Coding-guidelines#names). Of course we can come up with our own conventions and enforce them as strictly or as loosely as we'd like, but as  a default I'm leaning towards sticking to what's common in the community.

Agree. I've mostly been basing my ideal state there on what I'd seen previously and, looking at those docs, wonder if part of it might be related to the distinction between classes and functions. Although it seems like flagship also tended to follow (granted also inconsistent) what I had before of--UpperCamelCase functions that are exported, but not ones limited to a single file or internal to a class/function (regardless if made available to consumers).

I thought I read somewhere recently in React's docs of recommending only hooks having lower case function names, but can't readily find it.

Google is big and influential, I think I am fairly readily persuaded to follow that moving forward.

packages/components/src/components/Forms/FormText.tsx
```diff
@@ -0,0 +1,167 @@
...
+  return (
+    <Text aria-label={ariaLabel} style={textStyle}>
+      {getDisplayText(text)}
+      {required && <Text style={requiredStyle}>{` (*${t('required')})`}</Text>}
+    </Text>
+  )
+}
+
+/**
+ * Hint
```
TimRoe, line 81: I'm willing to let it slide as these are fairly closely tied to forms generally and not very multifunctional, but am noting that I'm not sure they strictly pass the `New functions have proper TSDoc annotations` checklist item.

TimRoe (changes requested):
That was a lot. 

Looks mostly good with mostly minor comments. 

Think the big request for changes is nixing the `Forms` folder to keep the components flat and instead shuffling the shared forms functions off to a new `utils` file.

packages/components/src/components/Forms/Checkbox/Checkbox.tsx
narin: Updated FormText elements to handle empty text values and return `null`

packages/components/src/types/common.ts
narin: Renamed `a11yLabel`

packages/components/src/utils/accessibility.ts
narin: Renamed function and use new a11yLabel field name

packages/components/src/components/Forms/FormText.tsx
narin: Moved Checkbox and CheckboxGroup back into components folder and created a new `shared` folder to contain FormText.tsx

packages/components/src/components/Forms/FormText.tsx
narin: Fixed

packages/components/src/components/Forms/CheckboxGroup/CheckboxGroup.tsx
narin: Added documentation and examples for onSelectionChange, selectedItems, and using TextWithA11yAndValue

packages/components/src/components/Forms/CheckboxGroup/CheckboxGroup.tsx
narin: Updated the comment for `value` to be a little clearer, but just to confirm, your main concern is less about the ability to provide `value`'s in the checkbox `items`, and more-so about whether this `TextWithA11yAndValue` type will be used across other form components?

I'm open to having them be local types to each CheckboxGroup and RadioButton, but this is kind of related to our coding style discussion. What should our criteria be for a type to be exported or in the shared type file like `forms.ts`? If it's used in at least X number of components? What is that number?

packages/components/src/types/forms.ts
narin: Moved it below and added it as a local prop type in CheckboxGroup

packages/components/src/types/common.ts
narin: Added `index.ts` file

packages/components/src/components/Forms/FormText.tsx
narin: Updated with more descriptive comments and comments above prop types

narin:
Resolved majority of comments. Added comment to continue discussion regarding `TextWithA11yAndValue` type. Will check in with @jessicawoodin regarding figma links.

packages/components/src/components/Forms/FormText.tsx
narin: Removed `export` to make it a local type

packages/components/src/components/Forms/CheckboxGroup/CheckboxGroup.tsx
TimRoe: > just to confirm, your main concern is less about the ability to provide `value`'s in the checkbox `items`, and more-so about whether this `TextWithA11yAndValue` type will be used across other form components?

Yeah. It's confusing to me as a global type since that implicitly means it should be useful to most components in my mind like the other types here which are relevant to basically every component that contains customizable text. As of this ticket, it's relevant to just CheckboxGroup and there's ambiguity in my mind if it does make sense to share 1:1 with Radio, let alone anything else after that.

> I'm open to having them be local types to each CheckboxGroup and RadioButton, but this is kind of related to our coding style discussion. What should our criteria be for a type to be exported or in the shared type file like `forms.ts`? If it's used in at least X number of components? What is that number?

Seems to me like `forms.ts` is a special case because the display logic itself is so repeatable, not sure we want to make a habit of abstracting typing too much. But to this ticket specifically: this is a global type currently that I'm not exactly clear should be shared at all so in my mind it should be local to the component and we can make it shared with the addition of Radio if it does make sense.

packages/components/src/components/shared/FormText.tsx
```diff
@@ -46,16 +47,19 @@ const fontDescription = {
 }
 
 /**
- * Header
+ * Header text element commonly used in form components
  */
```
TimRoe, line 51: Nitpicky, but this is aligned with the typing while others below are with the functional components.

packages/components/src/components/shared/FormText.tsx
```diff
@@ -0,0 +1,180 @@
+import { Text } from 'react-native'
+import { useTranslation } from 'react-i18next'
+import React, { FC } from 'react'
+
+import { StringOrTextWithA11y } from '../../types/common'
+import { getA11yLabel, getDisplayText } from '../../utils'
+import { useTheme } from '../../utils'
```
TimRoe, line 7: Imports can be merged

packages/components/src/components/CheckboxGroup/CheckboxGroup.tsx
```diff
@@ -0,0 +1,155 @@
+import { View, ViewStyle } from 'react-native'
+import { spacing } from '@department-of-veterans-affairs/mobile-tokens'
+import React, { FC, Fragment } from 'react'
+
+import { Checkbox } from '../Checkbox/Checkbox'
+import { ComponentWrapper } from '../../wrapper'
+import { Error, Header, Hint } from '../shared/FormText'
+import { FormElementProps } from '../../types/forms'
```
TimRoe, line 8: Can be merged with TextWithA11y import

TimRoe (approved):
Few minor suggested tweaks, but otherwise looks good--approved.

PR #537, opened 2024-10-09
[Tests] Checkbox, CheckboxGroup, FormText

narin:
## Description of Change

- Add unit tests for Checkbox, CheckboxGroup, and FormText
- Adds forgotten testID implemtnation to CheckboxGroup and Checkbox
- Checkbox and CheckboxGroup tests check for the existence of provided text while FormText tests for dark/light mode styling and accessibility labels.

#### Testing Packages

N/A

### Screenshots/Video

N/A

## Testing

 N/A

## PR Checklist
Code reviewer validation:
- General
	- [x] Acceptance criteria: 
		- All satisfied _or_
- Code

## Publish

Does not warrant publish

### DIFF
--- a/packages/components/jest/utils.ts
+++ b/packages/components/jest/utils.ts
@@ -1,0 +1,17 @@
+import { RenderResult } from '@testing-library/react-native'
+
+import { Icon } from '../src/components/Icon/Icon'
+
+/**
+ * Returns Icon on the screen
+ */
+export const getIcon = async (screen: RenderResult) =>
+  await screen.root.findByType(Icon)
+
+/**
+ * Returns the name of the Icon on the screen
+ */
+export const getIconName = async (screen: RenderResult) => {
+  const icon = await getIcon(screen)
+  return icon.props.name
+}
--- a/packages/components/src/components/Checkbox/Checkbox.test.tsx
+++ b/packages/components/src/components/Checkbox/Checkbox.test.tsx
@@ -3,7 +3,7 @@ import { fireEvent, render, screen } from '@testing-library/react-native'
 import React from 'react'
 
 import { Checkbox } from './Checkbox'
-import { getIcon, getIconName } from '../../../jest/utils'
+import { getIcon, getIconName } from '../../utils/jest'
 
 const mockedColorScheme = jest.fn()
 
--- a/packages/components/src/components/Checkbox/Checkbox.test.tsx
+++ b/packages/components/src/components/Checkbox/Checkbox.test.tsx
@@ -1,0 +1,155 @@
+import 'react-native'
+import { fireEvent, render, screen } from '@testing-library/react-native'
+import React from 'react'
+
+import { Checkbox } from './Checkbox'
+import { Icon } from '../Icon/Icon'
+
+const mockedColorScheme = jest.fn()
+
+jest.mock('react-native/Libraries/Utilities/useColorScheme', () => {
+  return {
+    default: mockedColorScheme,
+  }
+})
+
+describe('Checkbox', () => {
+  const onPressSpy = jest.fn()
+
+  const commonProps = {
+    header: 'Header text',
+    description: 'Description text',
+    hint: 'Hint text',
+    label: 'Label text',
+    testID: 'testCheckbox',
+    onPress: onPressSpy,
+  }
+
+  const errorMsg = 'Error text'
+
+  const getIconName = async () => {
+    const icon = await screen.root.findByType(Icon)
+    return icon.props.name
+  }
+
+  describe('Basic tests', () => {
+    it('initializes correctly', () => {
+      render(<Checkbox {...commonProps} />)
+      expect(screen.getByTestId('testCheckbox')).toBeTruthy()
+    })
+
+    it('renders header, description, hint, icon, and label text', async () => {
+      render(<Checkbox {...commonProps} />)
+      expect(screen.getByText('Header text')).toBeOnTheScreen()
+      expect(screen.getByText('Description text')).toBeOnTheScreen()
+      expect(screen.getByText('Label text')).toBeOnTheScreen()
+      expect(screen.getByText('Hint text')).toBeOnTheScreen()
+      expect(screen.queryByText('(*Required)')).not.toBeOnTheScreen()
+      expect(await getIconName()).toBe('CheckBoxOutlineBlank')
+    })
+
+    it('renders required indicator', async () => {
+      render(<Checkbox {...commonProps} required />)
+      expect(screen.getByText('(*Required)')).toBeOnTheScreen()
+    })
+
+    it('renders checked icon', async () => {
+      render(<Checkbox {...commonProps} checked />)
+      expect(await getIconName()).toBe('CheckBox')
+    })
+
+    it('renders indeterminate icon', async () => {
+      render(<Checkbox {...commonProps} indeterminate />)
+      expect(await getIconName()).toBe('IndeterminateCheckBox')
+    })
+
+    it('overrides checked if indeterminate', async () => {
+      render(<Checkbox {...commonProps} indeterminate checked />)
+      expect(await getIconName()).toBe('IndeterminateCheckBox')
+    })
+
+    it('fires onPress', async () => {
+      render(<Checkbox {...commonProps} />)
+      fireEvent.press(screen.getByRole('checkbox'))
+      expect(onPressSpy).toHaveBeenCalled()
+    })
+  })
+
+  describe('Styling', () => {
+    describe('Light mode', () => {
+      it('tiled variant styling (unchecked)', () => {
+        render(<Checkbox {...commonProps} tile />)
+        expect(screen.getByRole('checkbox')).toHaveStyle({
+          borderWidth: 2,
+          borderRadius: 4,
+          padding: 12,
+          paddingRight: 16,
+          borderColor: '#a9aeb1',
+          backgroundColor: '#ffffff',
+        })
+      })
+
+      it('tiled variant styling (checked)', () => {
+        render(<Checkbox {...commonProps} tile checked />)
+        expect(screen.getByRole('checkbox')).toHaveStyle({
+          borderWidth: 2,
+          borderRadius: 4,
+          padding: 12,
+          paddingRight: 16,
+          borderColor: '#005ea2',
+          backgroundColor: '#ecf1f7',
+        })
+      })
+
+      it('error state', () => {
+        render(<Checkbox {...commonProps} error={errorMsg} />)
+
+        expect(screen.root).toHaveStyle({
+          paddingLeft: 16,
+          borderColor: '#b50909',
+          borderLeftWidth: 4,
+        })
+      })
+    })
+
+    describe('Dark mode', () => {
+      it('tiled variant styling (unchecked)', () => {
+        mockedColorScheme.mockImplementation(() => 'dark')
+
+        render(<Checkbox {...commonProps} tile />)
+        expect(screen.getByRole('checkbox')).toHaveStyle({
+          borderWidth: 2,
+          borderRadius: 4,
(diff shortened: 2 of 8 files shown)


packages/components/src/components/Checkbox/Checkbox.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+    testID: 'testCheckbox',
+    onPress: onPressSpy,
+  }
+
+  const errorMsg = 'Error text'
+
+  const getIconName = async () => {
+    const icon = await screen.root.findByType(Icon)
+    return icon.props.name
+  }
```
TimRoe, line 33: Wonder if we should make a ticket for a utils file for test functions since, while we have two different variations (getIcon and getIconName), this is the 6th test suite using a generic getIcon function.

Not sure if we could similarly pull out color scheme mocking that every test uses, but might be worth exploring.

packages/components/src/components/Checkbox/Checkbox.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+          paddingRight: 16,
+          borderColor: '#005ea2',
+          backgroundColor: '#ecf1f7',
+        })
+      })
+
+      it('error state', () => {
+        render(<Checkbox {...commonProps} error={errorMsg} />)
+
+        expect(screen.root).toHaveStyle({
```
TimRoe, line 107: Maybe should also validate the presence of the error text since it doesn't seem other tests do, although that is a bit weird to fall under style tests.

packages/components/src/components/Checkbox/Checkbox.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+    })
+
+    it('fires onPress', async () => {
+      render(<Checkbox {...commonProps} />)
+      fireEvent.press(screen.getByRole('checkbox'))
+      expect(onPressSpy).toHaveBeenCalled()
+    })
+  })
+
+  describe('Styling', () => {
```
TimRoe, line 78: Probably also worth validating the colors on the checkbox icon.

packages/components/src/components/Checkbox/Checkbox.test.tsx
```diff
@@ -0,0 +1,155 @@
...
+      render(<Checkbox {...commonProps} />)
+      expect(screen.getByText('Header text')).toBeOnTheScreen()
+      expect(screen.getByText('Description text')).toBeOnTheScreen()
+      expect(screen.getByText('Label text')).toBeOnTheScreen()
+      expect(screen.getByText('Hint text')).toBeOnTheScreen()
+      expect(screen.queryByText('(*Required)')).not.toBeOnTheScreen()
+      expect(await getIconName()).toBe('CheckBoxOutlineBlank')
+    })
+
+    it('renders required indicator', async () => {
```
TimRoe, line 51: Wonder if this test should be more specific to validate against it being on label since the behavior is different for CheckboxGroup and in this manner it would pass just the same.

packages/components/src/components/shared/FormText.test.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+      render(<Error text={error} />, { wrapper: ComponentWrapper })
+      expect(screen.root).toHaveStyle({
+        color: '#fb5a47',
+      })
+    })
+  })
+
+  describe('Label', () => {
+    it('should render text and a11y label', () => {
```
TimRoe, line 101: Should validate the non-bold companion case to the subsequent test.

TimRoe: Ah, I see the test below does--still kind of weird to pair into the light mode test though.

packages/components/src/components/shared/FormText.test.tsx
```diff
@@ -0,0 +1,157 @@
...
+
+const mockedColorScheme = jest.fn()
+
+jest.mock('react-native/Libraries/Utilities/useColorScheme', () => {
+  return {
+    default: mockedColorScheme,
+  }
+})
+
+describe('Form Text', () => {
```
TimRoe, line 16: Seems like at least one set (potentially all of them?) should validate the non-object, string only scenario and that the a11yLabel is the same as the text.

packages/components/src/components/shared/FormText.test.tsx
```diff
@@ -0,0 +1,157 @@
...
+  const header = { text: 'Header text', a11yLabel: 'Header a11y' }
+  const hint = { text: 'Hint text', a11yLabel: 'Hint a11y' }
+  const description = {
+    text: 'Description text',
+    a11yLabel: 'Description a11y',
+  }
+  const error = { text: 'Error text', a11yLabel: 'Error a11y' }
+  const label = { text: 'Label text', a11yLabel: 'Label a11y' }
+
+  describe('Header', () => {
```
TimRoe, line 26: Should have test(s) for required true + special required styling.

packages/components/src/components/shared/FormText.test.tsx
```diff
@@ -0,0 +1,157 @@
...
+      mockedColorScheme.mockImplementation(() => 'dark')
+
+      render(<Error text={error} />, { wrapper: ComponentWrapper })
+      expect(screen.root).toHaveStyle({
+        color: '#fb5a47',
+      })
+    })
+  })
+
+  describe('Label', () => {
```
TimRoe, line 100: Should have test(s) for required true + special required styling.

packages/components/src/components/shared/FormText.test.tsx
TimRoe: Potentially also another should validate the object, but no special a11yLabel--to hedge against future scenarios when there's more than just text/a11yLabel on the object.

packages/components/src/components/CheckboxGroup/CheckboxGroup.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+      render(<CheckboxGroup {...commonProps} />)
+      const checkboxes = screen.queryAllByRole('checkbox')
+
+      // simulate uncheck '2'
+      fireEvent.press(checkboxes[1])
+      expect(onSelectionChangeSpy).toHaveBeenLastCalledWith(['Option 4'])
+    })
+  })
+
+  describe('Error state', () => {
```
TimRoe, line 101: Should also check error text presence.

packages/components/src/components/CheckboxGroup/CheckboxGroup.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+jest.mock('react-native/Libraries/Utilities/useColorScheme', () => {
+  return {
+    default: mockedColorScheme,
+  }
+})
+
+describe('CheckboxGroup', () => {
+  const onSelectionChangeSpy = jest.fn()
+
+  const items = [
```
TimRoe, line 18: One item should validate non-object, just string list item and corresponding behavior aligns as expected (text/a11y set to string, everything else not set).

packages/components/src/components/CheckboxGroup/CheckboxGroup.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+
+    it('renders required indicator', async () => {
+      render(<CheckboxGroup {...commonProps} required />)
+      expect(screen.getByText('(*Required)')).toBeOnTheScreen()
+    })
+
+    it('should mark selectedItems as checked', () => {
+      render(<CheckboxGroup {...commonProps} />)
+      const checkboxes = screen.queryAllByRole('checkbox')
+      expect(checkboxes[1].props.accessibilityState.checked).toBe(true)
```
TimRoe, line 72: Also be good to also validate the unchecked's.

packages/components/src/components/CheckboxGroup/CheckboxGroup.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+      expect(checkboxes[3].props.accessibilityState.checked).toBe(true)
+    })
+  })
+
+  describe('onPress behavior', () => {
```
TimRoe, line 77: Don't feel super strongly, but wonder if these would be improved by actually having a stateful render with spyOn and actually validating the behavior change vs. just using a mock function.

packages/components/src/components/CheckboxGroup/CheckboxGroup.test.tsx
```diff
@@ -0,0 +1,123 @@
...
+        'Option 4',
+        'Option 1',
+      ])
+    })
+
+    it('should pass updated selectedItems on uncheck', () => {
+      render(<CheckboxGroup {...commonProps} />)
+      const checkboxes = screen.queryAllByRole('checkbox')
+
+      // simulate uncheck '2'
```
TimRoe, line 95:
```suggestion
      // simulate uncheck '2', leaving just 'Option 4' selected
```

TimRoe (changes requested):
On the whole, looks good. Mostly minor comments, but requesting changes as there's a few functional gaps in testing.

packages/components/src/components/Checkbox/Checkbox.test.tsx
narin: Created https://github.com/department-of-veterans-affairs/va-mobile-library/issues/539

packages/components/src/components/shared/FormText.test.tsx
narin: Added non-bold test here and removed it from light mode color test

packages/components/src/components/shared/FormText.test.tsx
narin: Added non-object tests for each component and object without a11yLabel tests

packages/components/src/components/CheckboxGroup/CheckboxGroup.test.tsx
narin: Added additional tests to test array of strings

packages/components/jest/utils.ts
```diff
@@ -0,0 +1,17 @@
+import { RenderResult } from '@testing-library/react-native'
```
TimRoe, line 1: 1. Seems like the folder/file should be swapped? `utils/jest.ts` as it's somewhat similar to `storybook.tsx` of utils for a tooling set
2. Add to `.npmignore`

packages/components/src/components/shared/FormText.test.tsx
```diff
@@ -14,24 +14,49 @@ jest.mock('react-native/Libraries/Utilities/useColorScheme', () => {
...
+      expect(screen.root).toHaveTextContent('Header string')
+    })
+
+    it('should render text object without a11y label', () => {
+      render(<Header text={{ text: 'Header text object' }} />)
+      expect(screen.root).toHaveTextContent('Header text object')
+      expect(screen.getByLabelText('Header text object')).toBeOnTheScreen()
+    })
+
+    it('should render required text', () => {
```
TimRoe, line 53: In this test and/or another (the light/dark mode ones for red text?), I think it missed part of the earlier comment: "special required styling/a11y" to validate the red text of "(*Required)" and the a11y being `[text], Required`.

packages/components/src/components/shared/FormText.test.tsx
```diff
@@ -99,31 +148,44 @@ describe('Form Text', () => {
 
   describe('Label', () => {
```
TimRoe, line 149: Similar comment to above around "special required styling/a11y" case.

TimRoe (approved):
Approving, but a few minor tweaks recommended before merging:
- Moving jest utils file to utils folder (and adding to `index.ts` for easy import)
- Enhancing/adding tests to validate color/special a11y for the "(*Required)" text on Header/Label tests

